PR #108, opened 2013-01-25
fix the position when menu's css has a padding-top and padding-bottom

Leeiio:
If I add a padding-top and padding-bottom to .context-menu-list, the contextmenu position is wrong.


bbrala:
Made this change and merged the relevant code.

PR #120, opened 2013-02-24
Show CSS icons on HTML5 Polyfill

PawelStroinski:
Based on icon attribute of command element. This attribute contains icon path so use extracted file name without extension as an icon class (e.g. "images/cut.png" becomes "cut").


bbrala:
Has been added in #259

PR #121, opened 2013-02-26
Update src/jquery.contextMenu.js

dwaddell:
jQuery.andSelf has been deprecated: http://api.jquery.com/andSelf/

The solution is to use jQuery.addBack


Digicrat:
This issue (and fix) are still valid.

PR #152, opened 2013-06-22
twitter bootstrap compatibility

XAKEPEHOK:
Twitter bootstrap set display property as "inline-block" for .icon class


bbrala:
Made changed to how icons work to help developers use different icons. Also, removed the .icon class in favor for a namespaced icon class.

PR #161, opened 2013-07-29
added support for passing an id to add as a CSS selector

cthornbe:
When needing to do manipulation with CSS selectors in a project I am working on there was a need for an ID on the element. I believe this is something that could be useful for others in the same situation.


bbrala:
Id's should not be used lightly, dont feel this is needed.

PR #179, opened 2013-11-14
In html5 menus, use callback to compute disabled state to support dynami...

moesol:
...cly adding/removing disabled attribute.


src/jquery.contextMenu.js
```diff
@@ -1540,7 +1540,7 @@ function menuChildren(items, $children, counter) {
             case 'button':
                 item = {
                     name: $node.text(),
-                    disabled: !!$node.attr('disabled'),
+                    disabled: function() { !!$node.attr('disabled') },
```
bbrala: This will not work like this, no return

PR #190, opened 2014-01-30
Fixing hover issue

zyuhel:
Fixes hover issue when next first level items appear higher than third level ones of currently selected one.

### DIFF
--- a/src/jquery.contextMenu.css
+++ b/src/jquery.contextMenu.css
@@ -140,3 +140,9 @@
 .context-menu-accesskey {
     text-decoration: underline;
 }
+.context-menu-item {
+z-index: 1;
+}
+.context-menu-item.hover {
+z-index: 2;
+}


jzelenkov:
I've run a few tests and didn't notice this issue. @zyuhel in what browser / OS is this bug reproducible?

zyuhel:
Sorry, i easily reproduce it on our production, but couldn't find out what specifically leads to it. Because it works flawlessly on jsbin. Unfortunately i couldn't push all js/css code of SPA to jsbin to find out. So i think i should close the pull. I added this css to platform-fixes.css on production. It seems that to reproduce it very mad html structure is needed (all hail legacy). So i hope no one will stack upon it again.  
If i will find a clean way to reproduce it, i will reopen issue with demo. Sorry for it.

jzelenkov:
Cool. Thanks for your active participation!

PR #193, opened 2014-02-19
add compatibility with jquery ui icon

Roberto5:
with this edit we can add every icon that make with sprite, also jquery ui icon


bbrala:
This wont be merged. Made changed to how icons work to help developers use different icons.

PR #228, opened 2014-09-13
Changed .html() to .text()

gblock0:
The string passed to the menu items is eventually "eval"ed by jQuery, so using .html() allows XSS to get through and get executed by the browser.


awick:
So I was using this "feature" to do things like ```foo  <b>and not</b> bar```.  Is there a way to still do this?  Could a new type be added ( maybe "htmled" since html is already used) that has the old behavior?

PR #229, opened 2014-09-21
Fix for show-animation problem.

imcharsi:
Hi.

This is for show-animation problem.

https://github.com/medialize/jQuery-contextMenu/issues/157


bbrala:
This will break height calculations, sorry.

PR #231, opened 2014-10-18
Default Menu Item Text to Black

defcube:
The background is light grey, and in one of my sites the default text
color is white. This looked bad in the menu. Since the menu is supposed
to be general purpose, I believe this should be fixed in the core CSS.


bbrala:
I can understand it might not be nice visually, but until we might have an update the look i don't think this should be changed in the code. There might be people who styled this based on this color.

PR #235, opened 2014-11-18
playing with jQuery UI Theme integration

dev6:
Interesting!


bbrala:
I wont be picking this up. Use the new features for customization. :)

PR #243, opened 2015-01-26
Hide menu even if mouse down happens in another window (eg. iFrame)

dylangmiles:
I thought this could be useful to others so created a pull request.

I used the window.top.postMessage to basically notify any other open menus that a mouse down has occurred in another window so any open context menu can close itself.

I also did not want the overlay and took it out so that clicks can still select items underneath.

There seems to be an issue with Dalek visible and notVisible so two of the tests are broken.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 .DS_Store
 node_modules/
+.idea
+bower_components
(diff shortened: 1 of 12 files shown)


bbrala:
I start to see if this would work, but it seems there are still quite a lot of references to the overlay in the code. It doesn't work at the moment.

PR #246, opened 2015-02-26
Adds metroUI icons support

ToToSe:
Fix compatibility problem between metro ui icon and jQuery-contextMenu.
 
Conflicts between jQuery-contextMenu and metroUi Icon : 
Screen


bbrala:
Yeah the CSS naming kinda needs to be different, but the difference will break compatibility so shouldn't be done lightly really. If we do that it would mean an 2.0 release.

If that is the case i'd rather combine this with restructuring the way the css and such is generated so there is a little more flexibility.

bbrala:
Fixed in last merge. Icon classes are now namespaced.

PR #251, opened 2015-04-09
The context menu would incorrecly open another context menu, because it ...

NsdWorkBook:
...only looked at the trigger for the existing context menu, and not the one targeted

### DIFF
--- a/src/jquery.contextMenu.js
+++ b/src/jquery.contextMenu.js
@@ -226,6 +226,14 @@ var // currently active contextMenu trigger
             if (e.data.trigger != 'right' && e.originalEvent) {
                 return;
             }
+			
+			// Let the current contextmenu decide if it should show or not based on its own trigger settings
+            if (e.mouseButton !== undefined && e.data) {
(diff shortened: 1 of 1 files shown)


bbrala:
Ok cool, so basically if you have a context menu triggering on left mouse and one on right mouse this would fix opening both?

NsdWorkBook:
Yep that's the intention.

bbrala:
I tried to reproduce the the problem to write a test, but i wasnt able to. When i add a left click and a right click contextMenu to an element they behave like they shot. Do you have a testcase for me where the problem is appearant?

NsdWorkBook:
Ill try and make a test case for you.

bbrala:
Thank you :)

RareDevil:
I have made a small sample that shows this problem.
http://jsfiddle.net/RareDevil/c2ws3pc9/2/
If you left click the blue background then left click on one of the green boxes then that would open one of those.
you can also right click on one of the green boxes and then right click on the blue box for same result

bbrala:
Fixed in other pull request.

PR #256, opened 2015-05-15
fixes #252 add attribute hidden, allow options to be functions

### LINKED ISSUE #252: accesskey adds html as text
setting the accesskey option adds a span around the letter as text.

UziTech:


bbrala:
I've added the fix for #252 in another pull request. The hidden feature is implementet using visibility, would that do? See PR #258

bbrala:
Just reopen this request should there be anything else :)

PR #258, opened 2015-05-27
Added visible option to item

atipunov:
Fix copy/paste error in documentation
Added visible option to item
Improvements: if all options are hidden - don't show menu


bbrala:
Looks fine, ill just write a small test for the visibility.

Issue #240, opened 2014-12-23
Icon support in Importing HTML5 menu


robertobsc:
When importing html5 menu into jqueryContextMenu, it just ignores the icon tag, unlike the "normal" way of implementing the context menu.

In our case we create dynamic context menus in html5 way and pass to jqueryContextMenu to load it. Evertything works except the icon.

We started from this example in the site:
http://medialize.github.io/jQuery-contextMenu/demo/html5-import.html

example: < div id="menuContexto" class="context-menu-html5menu box menu-1">
< menu id="html5menu" type="context" style="display:none" class="showcase">
	< menuitem label="24046-3-45" onclick="javascript:detalharItemContextMenuById('transformador',556576);" icon="/gdis-ge-web/img/marcador.svg">< /menuitem>
	< menuitem label="Consultar Clientes" onclick="javascript:consultarClientes(556576);" icon="/gdis-ge-web/img/marcador.svg">< /menuitem>
	< menuitem label="Destacar Circuito" onclick="javascript:destacarCircuitoShowModal(556576);" icon="/gdis-ge-web/img/marcador.svg">< /menuitem>
	< menuitem label="Localizar" onclick="javascript:marcarTransformador(556576,true);" icon="/gdis-ge-web/img/marcador.svg">< /menuitem>
< /menu>
< /div >

bbrala:
Merged to master.

PR #259, opened 2015-06-06
Adding icon feature to html5 context menu

robertobsc:
I added this line at the code so we can create icons in html5 structure.

Would appretiate the merge!!

This solves the issue #240


bbrala:
Tests seem fine and should not really conflict.

PR #262, opened 2015-08-13
Fix for issue 197. I haven't looked into exactly why this works, but …

Dgc2002:
…work it does.

This is my first pull request, my apologies if I did something incorrectly.


bbrala:
Fixed in PR #265

PR #277, opened 2015-09-20
open contextMenu at 0,0

andreme:
The context menu was not opening when x or y were 0.


bbrala:
Good catch, thank you for the fix :)

PR #288, opened 2015-10-12
Do not place menu outside of client area

arai-a:
There is already similar code for offset.top. offset.left should also need it.


bbrala:
Thanks for the pull request, this should be fine to add. :)

PR #289, opened 2015-10-12
Update package.json to set jquery dep as peer dependency

fredericlb:
I'm having some issues of duplicated jQuery instances when using jQuery contextMenu with npm 3.x. This works great when jQuery is declared as a peer dependency (and this is the recommended declaration of dependencies for plugins).


bbrala:
I've read the docs, and agree on this one.  Thanks for the pull request! :)

PR #296, opened 2015-10-20
Redesign and fix for icon classname conflicts

RoachMech:
* added the 'context-menu' namespace to icon and state classes
* jQuery-contextMenu now uses sass to compile the css, which makes customization easier by using variables
* global redesign of the context menu
* ditched the image icon. jQuery-contextMenu now makes use of a generated iconfont

We need to credit https://github.com/encharm/Font-Awesome-SVG-PNG for using the svg icons.


bbrala:
@RoachMech whats left is maybe optimize the resulting CSS a bit, the icon classes are pretty bloated, that might not be needed if we add most of the css to its parent (context-menu-icon).

Also i need to check if the icon options are still ok like this :) Seems so, but still.

- [ ] Optimize resulting CSS for icons
- [ ] Check icon options in JS

RoachMech:
@bbrala please remove the context-menu-icon class from the generated dom

 [ ] remove context-menu-icon class

PR #299, opened 2015-10-25
Do not use .html() to markup accesskey

arai-a:
https://github.com/swisnl/jQuery-contextMenu/commit/83e37fbd3c6b0ca113597c01f03f62757cf86ecb changes .text() back to .html(), and it introduces the XSS risk again, that was once fixed by https://github.com/swisnl/jQuery-contextMenu/commit/f3137f21e17d3f91dad6f913eb3a0e93ada35b8a

accesskey's span element should be inserted separately from before/after text, to avoid parsing HTML there.

### DIFF
--- a/src/jquery.contextMenu.js
+++ b/src/jquery.contextMenu.js
@@ -1013,6 +1013,25 @@
 
                 root.accesskeys || (root.accesskeys = {});
 
+                function createNameNode(item) {
+                    var $name = $('<span></span>');
+                    if (item._accesskey) {
+                        if (item._beforeAccesskey) {
+                            $name.append(document.createTextNode(item._beforeAccesskey));
+                        }
+                        $('<span></span>')
+                            .addClass('context-menu-accesskey')
+                            .text(item._accesskey)
+                            .appendTo($name);
+                        if (item._afterAccesskey) {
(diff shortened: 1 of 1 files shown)


src/jquery.contextMenu.js
```diff
@@ -1042,7 +1061,13 @@
                         for (var i = 0, ak; ak = aks[i]; i++) {
                             if (!root.accesskeys[ak]) {
                                 root.accesskeys[ak] = item;
-                                item._name = item.name.replace(new RegExp('(' + ak + ')', 'i'), '<span class="context-menu-accesskey">$1</span>');
+                                var matched = item.name.match(new RegExp('^(.*?)(' + ak + ')(.*)$', 'i'));
+                                if (matched) {
+                                    item._beforeAccesskey = matched[1];
+                                    item._accesskey = matched[2];
+                                    item._afterAccesskey = matched[3];
+console.log(matched);
```
bbrala: Think you forgot a console.log here.

arai-a: Oops, thank you!
I'll fix it soon.

bbrala:
Wow, did i just catch you right here or something? Changing it in 2 minutes :)

PR #307, opened 2015-11-07
Remove executable bit from jquery.contextMenu.js

jacknagel:
This will propagate through to the files in `dist` the next time they are rebuilt.


bbrala:
Ok, sure :)

PR #313, opened 2015-11-11
Changed the defuat types. Now the span is inside the label

RareDevil:
I made this change so that is it possible to click the span with the text also to trigger the change event on inputs. Also this will make radio and check boxes be on the same line as the text instead of using two lines for those.

### DIFF
--- a/src/jquery.contextMenu.js
+++ b/src/jquery.contextMenu.js
@@ -1090,7 +1090,7 @@
                             $t.addClass('context-menu-html ' + root.classNames.notSelectable);
                         } else if (item.type) {
                             $label = $('<label></label>').appendTo($t);
-                            createNameNode(item).appendTo($t);
+                            createNameNode(item).appendTo($label);
 
                             $t.addClass('context-menu-input');
                             opt.hasTypes = true;


RareDevil:
Is there a reason that this have not been merged? I would like to know so if there needs to made some changes then its possible for me to do so.

bbrala:
I was looking into it, and to be honest, was too tired to oversee the possible rammifications.

I'll have a look again soon. :)

RareDevil:
Okay. thats fair enough, i just wondered about if this pull request had been overlooked.

RareDevil:
What are status on this one ?

bbrala:
Well, i looked at it again. And it seems the behaviour was changed in commit 873a63d478abb8c47deec16cd6440f6f6b84518d when changing how the name nodes are made. This actually fixes that :)

Thanks again, sorry it took a while to get this merged.

RareDevil:
Well that commit didnt fix this problem it acturelly made this problem. "createNameNode(item).appendTo($t);" This will add the NameNode to the item Container and that means that you cant click on the name node to change state of a radio / checkbox, you have to click on the radio / checkbox to change the state of them. Thats why i have made this commit that added the name node to the label so you are able to click the name node to change the state of radio / checkboxes. 

Also if you look close at line 1093 then you can see that there is this change:
$('<span></span>').html(item._name || item.name).appendTo($label);
becomes this
createNameNode(item).appendTo($t); 

.appendTo($label) -> .appendTo($t) - I see that as wrong.

bbrala:
Wow, i intended to merge this because i concluded you were right. So i updated the readme, but appearantly closed this pull request instead of merging it. Sorry.

RareDevil:
Oh okay that makes more sense. Sorry of i sounded a bit mad but it did not make sense that you closed this instead of merging it.

bbrala:
No worries :) I'm quite happy with your contribution, they help making the plugin better.

Issue #308, opened 2015-11-10
Submenu resize weirdly


RareDevil:
I have just updated to version 2.0.0 and noticed that the submenu sometimes resize weird and for some reason its missing 1 px in the width even tho the width is not over max-width. 
I have made a jsfiddle that shows the problem here: http://jsfiddle.net/RareDevil/vww70pLa/

I'm still trying to find out why this problem is there and i have not yet found the problem.

bbrala:
Hi, ill have a look this week to see what is happening. It seems it happens on the first item of submenu for me?

RareDevil:
I have updated my jsfiddle to show that is the longest of the items in the submenu that gets to two lines. http://jsfiddle.net/RareDevil/vww70pLa/1/ 
I have looked at the code and i wonder why is the width added by javascript and not managed with css ? Also i have looked at the sizing and it seems that every time its missing 1px, but i'm still working on pretty fix for this. 

I may be looking at making a pull request that removes the width from javascript and let css control it like i have managed to do it in my little project of learning to make jQuery widgets. I decided to make a contextmenu but that was just before swisnl took over this one. There is a sample of the menu here http://raredevil.ovh/contextmenu/

bbrala:
The width with JS is how it always worked, i'm not sure if that is really needed anymore, i'll as the designer @RoachMech to have a look.

In order to generate the new dist files for the CSS you need to have NodeJS and Npm installed. Using `npm install` in the root of the project should install everything you need to build the CSS from the sass files.

If you want to make changes to the CSS use the SASS files and regenerate the CSS with `gulp css`.

You can make the changes in https://github.com/swisnl/jQuery-contextMenu/blob/master/src/sass/jquery.contextMenu.scss, if you cant get gulp to run, that is fine too, ill regenerate the files when i accept the pull request.

RareDevil:
I have never used gulp so i just made the changes to the scss file it was just one small line.

bbrala:
I merged the SCSS and updated the dist files. I will walk to @RoachMech and ask him to look tomorrow, since he might've missed my ping.

RareDevil:
Sounds great. Cant wait to hear what will happen when @RoachMech looks at this.

RareDevil:
No news on this one ?

RoachMech:
I've looked into this and it seems like the entire resize() method is no longer of use and can therefore be removed. @bbrala can you confirm this?

[...]

PR #323, opened 2015-12-17
Properly fixed the problem with submenus size wrongly

RareDevil:
Using the css code i made before was wrong. I found out that if you changed the $menu.width() to $menu.outerWidth() it would get the correct size.

This pull request can close the issue #308 problem solved and fixed.


bbrala:
Checked and merged :+1:

PR #326, opened 2015-12-18
Added support for providing a function as zIndex value in options obj…

eivindga:
Use case: 
- in applications where you have to calculate a dynamic zIndex, it would be nice to be able to provide a function for calculating the zIndex. 

For instance if the click target has zIndex 1, and a jquery dialog right below has zIndex 347 (just some variable number), you would like to have the context menu displayed in front of the dialog.

### DIFF
--- a/src/jquery.contextMenu.js
+++ b/src/jquery.contextMenu.js
@@ -268,7 +268,7 @@
                 if ((e.data.trigger !== 'right' && e.data.trigger !== 'demand') && e.originalEvent) {
                     return;
                 }
-				
+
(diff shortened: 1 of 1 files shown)


bbrala:
Thanks for the contribution, the code seems fine. I'll build the dist for you. Also, it seems you committed the file with a change to the access bits (src/jquery.contextMenu.js 100644 → 100755), i'll fix that too.

bbrala:
I also changed the argument so it gets the same arguments as other callbacks. :)

eivindga:
Great! :)
And sorry for the access bit changes, guess that was just something Windows thought was a bright idea...

PR #329, opened 2015-12-25
Positioning of contextmenu when using appendTo

mrMarco:
When a selector has been assigned to the appendTo option, the context menu wasn't (always?) positioned at the cursor position.
This fix solves the problem.

### DIFF
--- a/src/jquery.contextMenu.js
+++ b/src/jquery.contextMenu.js
@@ -332,7 +332,10 @@
                     }
                     if (showMenu) {
                         // show menu
-                        op.show.call($this, e.data, e.pageX, e.pageY);
+		                var menuContainer = (e.data.appendTo === null ? $('body') : $(e.data.appendTo));
+		                op.show.call($this, e.data,
+                					 $(e.srcElement).offset().left - menuContainer.offset().left + e.offsetX,
+                					 $(e.srcElement).offset().top - menuContainer.offset().top + e.offsetY);
                     }
                 }
             },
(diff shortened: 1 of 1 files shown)


bbrala:
In the tests it breaks on what seems a lack of srcElement. I think srcElement is not the cross browser property for that.  Should this just be .target?

mrMarco:
@bbrala: You're right. that isn't browser-independant. I've changed the new lines slightly to solve that. The new lines are now like this:

var menuContainer = (e.data.appendTo === null ? $('body') : $(e.data.appendTo));
var srcElement = e.target || e.srcElement || e.originalTarget;
op.show.call($this, e.data,
                    $(srcElement).offset().left - menuContainer.offset().left + e.offsetX,
                    $(srcElement).offset().top - menuContainer.offset().top + e.offsetY);

src/jquery.contextMenu.js
```diff
@@ -332,7 +332,11 @@
                     }
                     if (showMenu) {
                         // show menu
-                        op.show.call($this, e.data, e.pageX, e.pageY);
+		                var menuContainer = (e.data.appendTo === null ? $('body') : $(e.data.appendTo));
+		                var srcElement = e.target || e.srcElement || e.originalTarget;
```
bbrala: This can be .target since jQuery normalizes the event.

bbrala:
Thanks for the contribution, it has been merged.

PR #341, opened 2016-01-13
Fixed a problem when using the open function

RareDevil:
I had a problem where e.offsetX and e.offsetY is undefined. This was mainly when i used the open function like the example below.
`$(selector).contextMenu({ x: ShowX, y: ShowY });`
So i have made a check if they are undefined and if they are pageX and pageY are used instead.

### DIFF
--- a/src/jquery.contextMenu.js
+++ b/src/jquery.contextMenu.js
@@ -339,9 +339,7 @@
                                          $(srcElement).offset().left - menuContainer.offset().left + e.offsetX,
                                          $(srcElement).offset().top - menuContainer.offset().top + e.offsetY);
                         } else {
-                            op.show.call($this, e.data,
-                                         $(srcElement).offset().left - menuContainer.offset().left + e.pageX,
-                                         $(srcElement).offset().top - menuContainer.offset().top + e.pageY);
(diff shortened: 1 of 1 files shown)


RareDevil:
@bbrala Have the time to look at this?

bbrala:
The first thing that i was thinking is that ShowX and ShowY aren't documented so shouldn't you use other properties to do this?

http://swisnl.github.io/jQuery-contextMenu/docs/events.html

RareDevil:
Well ShowX and ShowY is just a name i used instead of typing in a random number like this
`$(selector).contextMenu({ x: 50, y: 25 });`
But the main thing was that when i used this command i experienced that e.offsetX and e.offsetY was undefined in the code. If you try to look at code i changed then before my fix it used  e.offsetX and e.offsetY and those was undefined so all i did was made a check to make sure that they was not undefined before they was used and if they is undefined e.pageX, e.pageY is used instead.

bbrala:
Interestingly the docs even say: ```// or $('.context-menu-one').contextMenu({x: 100, y: 100});```

I've merged your change, thanks again for the contribution :)

PR #343, opened 2016-01-21
Simulate click on the element if the menu already open

nextend:
User opens the menu with right click, then he right click again on a different part, contextMenu should inform that "clicked, but hidden by $layer" element about the click event. 

In our system, we have one main contextMenu and our system contains an active item which can be selected by click. If this event not triggered, then the second right click not able to change the active item to the clicked one.


bbrala:
Thanks for the contribution. I looked through the code, and this would trigger mouse events that might have been bound with jQuery. I'm thinking this might result in also triggering the target element for the menu.

So there might need to be a check around that, i need to have a closer look to be sure.

nextend:
Maybe it could be added as an option, so it wouldn't break the original code, but the improvement would be available in the future.

PR #354, opened 2016-02-22
Bugfix: fixed positioning context menu when used with d3 tree

mehranshakeri:
Using jQuery contextMenu with d3 tree (http://bl.ocks.org/mbostock/4339083) object creates a problem with positioning the menu in wrong place.

ContextMenu in opening time checks "offsetX" attribute of event and in case if exists, it does some calculation to position the menu. But in combination with d3 tree, this value will be calculated wrong and menu will be positioned in far right side of page. I have seen in absence of "offsetX" attribute, "pageX" is used. I could fix this issue with replacing the whole condition block with just else part. 
Is there any reason for using offsetX when pageX exists all the time?

### DIFF
--- a/src/jquery.contextMenu.js
+++ b/src/jquery.contextMenu.js
@@ -337,13 +337,7 @@
                         // show menu
 		                var menuContainer = (e.data.appendTo === null ? $('body') : $(e.data.appendTo));
 		                var srcElement = e.target || e.srcElement || e.originalTarget;
-                        if (e.offsetX !== undefined && e.offsetY !== undefined) {
-                            op.show.call($this, e.data,
-                                         $(srcElement).offset().left - menuContainer.offset().left + e.offsetX,
-                                         $(srcElement).offset().top - menuContainer.offset().top + e.offsetY);
-                        } else {
-                            op.show.call($this, e.data, e.pageX, e.pageY);
-                        }
+                    op.show.call($this, e.data, e.pageX, e.pageY);
                     }
                 }
             },
@@ -521,7 +515,7 @@
 					} else {
 						if (target.offsetParent !== null && target.offsetParent !== undefined) {
 							getZIndexOfTriggerTarget(target.offsetParent);
-						} 
+						}
 						else if (target.parentElement !== null && target.parentElement !== undefined) {
 							getZIndexOfTriggerTarget(target.parentElement);
 						}


anseki:
It is positioned wrong, in pages without D3 also.
See this:
https://jsfiddle.net/fhmf00oy/

In Chrome, I right clicked "TRIGGER" and I didn't move mouse:

s01

`offsetX/Y` still have compatibility problems.
See this:
https://jsfiddle.net/ekbj8asf/

Firefox:
s02

Chrome:
s03

It seems Chrome gets those of inline-element by `content-box`-based, and Firefox gets those by `border-box`-based.

As @mehranshakeri said, I also think that `pageX/Y` should be used.

anseki:
This issue occurs in IE also.

bbrala:
Thanks for the contribution. I'll have a look at it soon. I'll try and find out why there is a offsetX used so we might be able to remove it. The pageX fallback when there is no offsetX has been added only this januari, so i'll dig around and see if i can see a valid reason to keep it.

alihb2005:
By putting the "e.pageX" and "e.pageY" in the last block of the "contextMenu" function for the call to "show" funtion the problem will be solved

AllBogs:
Howdy. I seem to be running into the same issue in a project. As a workaround, I have reverted to version 2.0.1 for the time being.

servocoder:
Same thing!
It worked fine for me untill I added `div` wrapper to the page to center a content as:
```
<body>
    <div class ="wrapper">
        Content here...
    <div>
</body>
```
```
.wrapper {
    width: 1200px;
    margin: 0 auto;
}
```

I have reverted to 2.0.1 as @AllBogs did, and it works correct.
Fix please.

bbrala:
Hi guys,

I was looking into this and my conclusion is the earlier "fix" that added offsetX should be reverted sadly. I'll try and accept these changes this week and release a new version so you update again.

Thanks all for the input.

anseki:
At least, @mehranshakeri's patch seems to working fine.

bbrala:
Yeah his patch is basically remove the offsetX/Y from the code again :)

bbrala:
Note to self; original PR #354

bbrala:
I merged this to fix the issue for the people having it. 

#341 needs to be revisited.

Issue #358, opened 2016-02-27
Submenus are not collapsed when the menu is closed


anseki:
Hi, thank you for great plugin.

So, I feel behavior of submenu is strange when it is shown again.
If a menu is closed before the submenus are hidden, those submenus are shown without being selected when the menu is opened next time.
https://swisnl.github.io/jQuery-contextMenu/demo/sub-menus.html

![01](https://cloud.githubusercontent.com/assets/4763469/13373424/838af02c-ddab-11e5-879a-22295761d395.gif)

And also, nested submenus are shown when its parent item was focused even if the menu is not closed:

![02](https://cloud.githubusercontent.com/assets/4763469/13373429/938a9450-ddab-11e5-8089-4dfd959d0f0e.gif)

And also, those submenus are still focused.

I think that all submenus should be collapsed when the menu is closed (whether a menu item is clicked or it is canceled) or its parent item is unfocused.

`hide` operation doesn't hide submenus.
`focusItem` hides only submenus of siblings.
`itemMouseenter` and `focusItem` don't make any item blur because [`children`](https://github.com/swisnl/jQuery-contextMenu/blob/master/src/jquery.contextMenu.js#L803) and [`filter`](https://github.com/swisnl/jQuery-contextMenu/blob/master/src/jquery.contextMenu.js#L889) are passed a class name, these require a selecter (i.e. `.` + class name).

bbrala:
Thank you for this lovely bug report. Looked at the pull request, that's indeed an oversight from when i changed the classnames to be dynamic a while ago. Good catch.

PR #359, opened 2016-02-27
Fix: Submenus are not collapsed when the menu is closed (#358)

anseki:
- Remove `classNames.visible` from all items in `hide` operation
- Make other items blur in `focusItem` and `itemMouseenter`


bbrala:
I've merged this manually, thanks a lot :+1:

Issue #360, opened 2016-02-27
`width` is increased


anseki:
When a menu is opened, `width` of its element is increased every time.
This issue was reported sometimes in the past, but it's not fixed yet. https://github.com/swisnl/jQuery-contextMenu/issues/197 , https://github.com/swisnl/jQuery-contextMenu/issues/183
This occurs in official demo page also:
https://swisnl.github.io/jQuery-contextMenu/demo.html
I right clicked and pushed escape key, repeatedly:
![01](https://cloud.githubusercontent.com/assets/4763469/13373695/faee8b5a-ddb1-11e5-832e-c338b1629f4d.png)

And also, if the menu has sub menus, those elements are overlapped because a sub menu has another [issue](https://github.com/swisnl/jQuery-contextMenu/issues/358):
![02](https://cloud.githubusercontent.com/assets/4763469/13373706/301a0f84-ddb2-11e5-9516-346255d14231.png)

A got `outerWidth` that includes `border-width`, `padding`, etc. at [here](https://github.com/swisnl/jQuery-contextMenu/blob/master/src/jquery.contextMenu.js#L1290) is set to `width` that doesn't include those at [here](https://github.com/swisnl/jQuery-contextMenu/blob/master/src/jquery.contextMenu.js#L1309).
That is, `border-width` (and `padding` if styles have it) are added to `width` every time.

PR #361, opened 2016-02-27
Fix:  `width` is increased (#360)

anseki:
Adjust size of element via `outerWidth` in `resize`.


bbrala:
Thanks again for the great report en PR. :+1: i merged it manually again.

PR #363, opened 2016-03-01
Disabled items styles

anseki:
Hi,
This PR changes styles of disabled items.

- Now, a mouse cursor on disabled items is "I-beam". I think that it should be arrow cursor.
See mouse cursor in this screen:

s-01

It seems that this line https://github.com/swisnl/jQuery-contextMenu/blob/master/src/sass/jquery.contextMenu.scss#L81 means that, but `context-menu-hover` class is never set to the items because the disabled items are not focused. https://github.com/swisnl/jQuery-contextMenu/blob/master/src/jquery.contextMenu.js#L806
So, I think that `context-menu-hover` class is not necessary to a mouse cursor because `cursor` property is set to styles of the item (i.e. `cursor` property works only when a mouse cursor is hovering on the item).

- I think that icons also should be gray when those are disabled.
- It seems that line https://github.com/swisnl/jQuery-contextMenu/blob/master/src/sass/jquery.contextMenu.scss#L71 and https://github.com/swisnl/jQuery-contextMenu/blob/master/src/sass/jquery.contextMenu.scss#L74 are duplicated.
- I feel that the gray color is almost black. It should be more gray like other apps.

Screen after change by this PR:

s-02

(Sorry my English is very poor)


anseki:
I checked other apps more, and I changed the color to make users understand disabled items even when icons are not shown.
s-03

bbrala:
Thanks for thew contribution, i agree the default disabled style can use some work. I'll have a look at this soon.

PR #364, opened 2016-03-05
Reduced size

anseki:
Hi,
I found that width of the menu is reduced in some cases.
For example, `items`:

```js
items: {
  item1: { name: 'foo bar foo bar foo bar' },
  item2: { name: 'foo bar foo bar foo bar' },
  item3: { name: 'foo bar foo bar foo bar 1' },
  item4: { name: 'foo bar foo bar foo bar' },
  item5: { name: 'foo bar foo bar foo bar' }
}
```

s1

The 3rd item was split to 2 lines.

Note that this depends on the environment such as font-family, font-size, etc,. and OS and browser also maybe. If this is not reproduced in your computer, try to adjust those.
This occurs in current version of Chrome, Firefox and IE in my computer at least. `font-family: Verdana` and `font-size: 16px` were shown in Dev Tools.

The cause is that the `outerWidth` method returns rounded value as pixels.
I think that `Math.ceil` is used to avoid that the width is reduced, in this line, right?:
https://github.com/swisnl/jQuery-contextMenu/blob/master/src/jquery.contextMenu.js#L1292
But, that was already rounded. It seems that this behavior of jQuery is still being debated. https://github.com/jquery/jquery/issues/1724

Anyway, the exact value can be got by `getBoundingClientRect` method. And, for browsers that don't support `getBoundingClientRect`, 1px is added to width in this PR. Is there any problem about this 1px?


bbrala:
Thanks for the pull request! 

From what i remember there might be an issue with changing the way the width is checked since it is done in more than one place, we had some issues earlier with the width of the menu  growing when reopening. I'll try and have a look at this pull request soon to see if this can be merged safely.

Issue #302, opened 2015-10-26
Support HTML as content for the name of an item


bbrala:
As discussed in #252 a way to add HTML as an item that plays nicely with the different item types.

arai-a:
there could be 2 ways to solve it:

1. Different property for HTML
  If `item.html` property exists, call `.html(item.html)` instead of `.text(item.name)`

2. Flag property to mark it as HTML
  If `item.isHTML` property is true, call `.html(item.name)` instead of `.text(item.name)`

in both way, property name is just a draft, there could be better name.

The problem is that there is already `item.type == 'html'` with `item.html` property for content.  It should be nice to make these 2 types clearly distinguishable.

brassard:
Proposed a PR #388 to address this issue.

brassard:
I believe that this issue is resolved with the now-merged PR #388.

bbrala:
Yes! :) I went through the issues, but didnt go back enough. Thanks

PR #370, opened 2016-03-11
fix #369

### LINKED ISSUE #369: Allow html name
Hello,
Since 2.0 version if we defined an item with a name wich contain html, the html tags was not interpreted.

Exemple:

if we define name like this : `&eacute;t&eacute;` , the generate name is `&eacute;&eacute;`  not été.

For fix it I change the function "createNameNode" on ligne 1078, I replace ".text(...)" by ".html(...)" 

```
function createNameNode(item) {
    var $name = $('<span></span>');
    if (item._accesskey) {
        if (item._beforeAccesskey) {
            $name.append(document.createTextNode(item._beforeAccesskey));
        }
        $('<span></span>')
            .addClass('context-menu-accesskey')
            .html(item._accesskey)
            .appendTo($name);
        if (item._afterAccesskey) {
            $name.append(document.createTextNode(item._afterAccesskey));
        }
    } else {
        $name.html(item.name);
    }
    return $name;
}
```

zendre4:


bbrala:
As mentioned in the issue, i can't merge this. We need to find another way to support this.

zendre4:
Ok no problem. I will follow the #302 issue! Thank for response

PR #371, opened 2016-03-14
Fix: Disabled items get focus

anseki:
Disabled items that have sub-menu are focused by accesskey.

For example:

```js
items: {
  itemA: {
    disabled: true,
    name: 'A foo bar',
    accesskey: 'a',
    callback: function(key) { console.log('key: ' + key); }
  },
  itemB: {
    name: 'B foo bar',
    accesskey: 'b',
    callback: function(key) { console.log('key: ' + key); }
  },
  itemC: {
    name: 'C foo bar',
    accesskey: 'c',
    callback: function(key) { console.log('key: ' + key); }
  },
  itemD: {
    disabled: true,
    name: 'D foo bar',
    accesskey: 'd',
    items: {
      sub1: { name: 'sub 1' },
      sub2: { name: 'sub 2' },
      sub3: { name: 'sub 3' }
    },
  },
  itemE: {
    name: 'E foo bar',
    accesskey: 'e',
    callback: function(key) { console.log('key: ' + key); }
  },
  itemF: {
    disabled: true,
    name: 'F foo bar',
    accesskey: 'f',
    callback: function(key) { console.log('key: ' + key); }
  }
}
```

`itemA` and `itemF` ignore focus correctly.
`itemD` also ignore focus when mouse cursor is hovered, but it is focused when `D` key is pressed.

s01

When an accesskey is pressed, an event handler is chosen:
https://github.com/swisnl/jQuery-contextMenu/blob/master/src/jquery.contextMenu.js#L655
The event handler `itemClick` checks `disabled` of the item, and it ignores disabled item:
https://github.com/swisnl/jQuery-contextMenu/blob/master/src/jquery.contextMenu.js#L833
But, the event handler `focusItem` does not it:
https://github.com/swisnl/jQuery-contextMenu/blob/master/src/jquery.contextMenu.js#L877


bbrala:
Thanks for the contribution and very good description! I will have a look soon, but on a first glance this seems fine indeed.

PR #379, opened 2016-04-06
Prevent recreation or reposition happening on mousedown

### LINKED ISSUE #332: Recreation/reposition is activated on button down instead of button up
The context menu is opened on mouse button up which is consistent with what most UIs do.

However, once a menu is opened, pressing the mouse button again on another triggering element the cancel and the recreation both happen on trigger down which is inconsistent with the original creation of the menu.

To be consistent, the cancel should happen on trigger down but the recreation/reposition should happen on trigger up.

I think it would help with the feel of the menu.

Thanks.

Kiedro:
Fix #332 . Removing recreation/reposition on mousedown is more consistent and also prevent users from miss clicking menu option in case pointer slightly moves into menu beetwen mousedown and mouseup what is a common case.

### DIFF
--- a/src/jquery.contextMenu.js
+++ b/src/jquery.contextMenu.js
@@ -435,62 +435,12 @@
             // click on layer to hide contextMenu
             layerClick: function (e) {
                 var $this = $(this),
-                    root = $this.data('contextMenuRoot'),
-                    button = e.button,
-                    x = e.pageX,
-                    y = e.pageY,
-                    target,
-                    offset;
+                    root = $this.data('contextMenuRoot');
 
                 e.preventDefault();
                 e.stopImmediatePropagation();
 
                 setTimeout(function () {
-                    var $window;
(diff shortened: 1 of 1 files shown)


bbrala:
In your description you say you fixed a discrepance, but in de code i only really see you removing the code that decides to move the menu instead of recreating it?

Kiedro:
What I tried to achieve:
when a menu is open and user performs a right click, menu is hidden on a mousedown, and then if mouseup happens on triggering element menu is opened. So there is not visible menu in between mousedown and mouseup.

I'm not sure if it's possible to reposition menu and hide it, so the recreting is not necessary, if yes then my pr is wrong. 
Maybe the whole option for repositioning instead of recreating should be dropped?

bbrala:
This PR would break a lot since it just removes functionality. Sorry, but closing this one.

Issue #380, opened 2016-04-08
build-icons character set on windows


roselan:
when I build icons on a windows machine, all icons are displayed the "missing char in font" " empty box.

What a tried: 
* build icons
* get svg icons from the svg-pgn-font-wesome rep, build icons
* got all gulp deps, and gulped full project.
* got original woff, eot etc font files from the font-awsom rep, and tried to hack them in contextMenu.

No strawberry.

I did notice that in the sassed css, icon characters do not correspond to the original when i build it through gulp. By example, "delete" icon character becomes "EE A8 84" instead of "EE AA BC".

I'm on a windows machine, and I found a gulp issue that might explain the behavior:

I pointed it to the gulp team too: https://github.com/gulpjs/gulp/issues/1546#issuecomment-207314603

bbrala:
Hmm, weird, i've had no issues, but i build on Linux. To make things a little more visible i made a branch where the icon character is a readable string in commit e0b8f7d3e563f5073d7db1c526ae4ba51e10d5dd.

roselan:
Thank you. 

That's exactly what I did (well, nearly, i added the backslash directly:` "\<%= glyph.unicode[0].charCodeAt(0).toString(16).toUpperCase() %>"`)

But that didn't work as expect. by example: you generate 
```
.context-menu-icon-delete:before {
  content: "\EA04";
}

```
and I generate:
```
.context-menu-icon-delete:before {
  content: "\EABC";
}
```

**However**. That's with all svg fonts from https://github.com/encharm/Font-Awesome-SVG-PNG.

When I build-icons with only the ones you provided, it works!

both your and and encharm delete.svg files are strictly equals.

I'm quite certain the fault is on me, it's my first time using npm/gulp. Please note that when I build, i get the 4 warning, and have to interrupt it (but it works). I did copy all encharm icons inside the "icon" folder and ran:

```
C:\projects\jQuery-contextMenu-master>gulp build-icons
[19:42:36] Using gulpfile C:\projects\jQuery-contextMenu-master\gulpfile.js
[19:42:36] Starting 'build-icons'...
[19:42:37] Finished 'build-icons' after 680 ms
[19:42:37] gulp-svgicons2svgfont: Font created

C:\projects\jQuery-contextMenu-master>gulp
[19:42:47] Using gulpfile C:\projects\jQuery-contextMenu-master\gulpfile.js
[19:42:47] Starting 'jshint'...
[...]

PR #381, opened 2016-04-08
force font woff2 creation

roselan:
#380 On windows at least, the woff2 font file is not created, and without it, icons are not displayed.


bbrala:
Thanks :) Guess it wasn't being regenerated. Oops.

PR #382, opened 2016-04-09
Correct typos / mistakes in the main documentation page

xiatianyu:
➊ Unlike the jQuery ID selector, the correct syntax for `document.getElementById()` does not use `#` in its argument string.
➋ In the value section of `trigger`, `appendTo `was mistakenly used in place.
➌ In the 2nd example for `trigger`, the first line stores the `context-menus-container` DOM element inside a variable, but it serves no purpose later on.

### DIFF
--- a/documentation/docs.md
+++ b/documentation/docs.md
@@ -89,7 +89,7 @@ $.contextMenu({
 });
 
 // select the container with a dom element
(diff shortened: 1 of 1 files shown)


bbrala:
Thank you for the correction!

PR #388, opened 2016-04-19
adding option for menu item name to use html() instead of text()

brassard:
After upgrading to 2.x, one thing that has changed since 1.8.1 is the ability to put styled HTML into the names of entries. It is useful to allow for the use of this, even though it could have a potential security risk for dynamically named items.

### DIFF
--- a/src/jquery.contextMenu.js
+++ b/src/jquery.contextMenu.js
@@ -1089,7 +1089,11 @@
                             $name.append(document.createTextNode(item._afterAccesskey));
                         }
                     } else {
-                        $name.text(item.name);
+                        if (item.isHtmlName) {
+                            $name.html(item.name);
+                        } else {
+                            $name.text(item.name);
+                        }
                     }
                     return $name;
                 }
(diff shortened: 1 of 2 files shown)


src/jquery.contextMenu.js
```diff
@@ -1089,7 +1089,11 @@
                             $name.append(document.createTextNode(item._afterAccesskey));
                         }
                     } else {
-                        $name.text(item.name);
```
brassard: Would we need to make a similar change around `jquery.contextMenu.js:1086`?

bbrala: I think this might make things weird. Combining the HTML with accessKey will need to make sure the accessKey is not in the HTML parts of the name. Or the decision should be they cant exist together. I'm not entirely sure what is the best approach right now.

brassard: Yea... I'd say that it would be a reasonable approach to "disable" the accessKey functionality altogether if the name is defined as being HTML.

bbrala:
Could you add the documentation for this feature in [documentation/docs/items.md](https://github.com/swisnl/jQuery-contextMenu/blob/master/documentation/docs/items.md)?

brassard:
I added some documentation and some error handling. See what you think.

bbrala:
Thanks for the contribution! :)

PR #395, opened 2016-05-20
Remove deprecated option in 'bower.json'.

nelson6e65:
`version` option is _deprecated_ and is ignored by **Bower**, as mentioned in its [specification](https://github.com/bower/spec/blob/master/json.md#version).


bbrala:
Thanks for the contribution :)

Issue #352, opened 2016-02-17
Cannot read property 'trigger' of undefined


andreasrosdal:
[Freeciv-web](https://github.com/freeciv/freeciv-web) uses jQuery contextMenu v2.1.0 to show the context menu. We use a JavaScript error logging-tool called Track:js, which has reported these JavaScript errors in contextMenu.

Reported 897 times by Track:js:
![trackjs-1](https://cloud.githubusercontent.com/assets/210993/13122350/c6a93c6c-d5b6-11e5-9e43-6d187aae3a16.png)

Reported 283 times:
![trackjs-2](https://cloud.githubusercontent.com/assets/210993/13122352/c8e9785c-d5b6-11e5-8d9d-dded4b1f5004.png)

Any help in resolving these?

andreasrosdal:
This seems related to bug #351.

bbrala:
Thanks for reporting this, i'll try and make time this week to go through this issue.

jasekiw:
I also have ran into this problem. It seems to only happen when using anonymous callbacks in the same scope. A work around is to set a variable to trigger and then use that instead.

`var target = opt.$trigger;`

bbrala:
@jasekiw Do you mean on the items? I've been trying to reproduce this problem, but i can't really at the moment. Do you have a configuration that breaks for me?

PR #416, opened 2016-07-05
Add checks for null before using opt.$menu and root.$menu.

andreasrosdal:
This is a pull-request for issue #352 which I've reported eariler. I used to get quite a lot of Javascript errors of this kind reported on play.freeciv.org by Trackjs (error reporting tool).

I have been using this change in production on play.freeciv.org for about one week, and it has resolved all these javascript errors from being reported any more.

### DIFF
--- a/src/jquery.contextMenu.js
+++ b/src/jquery.contextMenu.js
@@ -485,7 +485,7 @@
                         });
                     }
 
-                    root.$menu.trigger('contextmenu:hide');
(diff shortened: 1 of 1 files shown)


bbrala:
Thanks for the contribution. Still puzzled when this happens. But some extra checks shouldn't introduce problems i think. I will merge this in soon and regenerate the dist files.

PR #417, opened 2016-07-05
Change second 'show' to 'hide' in events docs

kgeorgiou:
`show` is used twice, I _assume_ from the context that the second `show` should actually be a `hide`.


bbrala:
Yeah you are right, thank you for the contribution. Will merge this soon.

PR #428, opened 2016-07-17
Improve docs on icon additions

ElDeveloper:
I found this while trying to add some custom icons.


bbrala:
Thanks for the contribution! 

So ``gulp build-icons`` was not enough to generate the correct CSS? That is what i thought.

ElDeveloper:
Not a problem! :tada:

Yes, it took me a little while to figure out exactly what was going on
i.e. I added a few icons and couldn't get them to work by just replacing
the font files.

BTW, speaking about fonts, is there a particular font file I should be
using or should I use the four of them i.e. woff, woff2, eot and ttf?

Issue #429, opened 2016-07-19
Asynchronous create sub menu's after context menu has opened.


Ruud-cb:
Hi,

I've been searching the past 2 days for a perfect context menu, this one and another one stand out above all others. To put it in perspective: I am developing an app with AngularJS. Thus I rather have a angular context menu then a default jquery menu.
One came really close: https://github.com/Templarian/ui.bootstrap.contextMenu
Yet this one was not so nice with providing icons, keyboard events some other things. But it does has one major improvement above this one: Async create sub menu's. Instead of providing an item I can provide an [promise](https://api.jquery.com/promise/), which would be loaded once the user has opened the sub-menu.

Perhaps it would be possible to implement something similar to this? So instead of only providing items, or a list of items, provide a promise, and once it is completed it will render the sub menu's. 

I know the documentation provides something with async loading the context menu, but that's not what I am looking for, that happens before everything is created. This should happen once the context menu has already been created, either after the menu has opened, or once the user wants to open the sub menu.

See the plunker for an example using that angular directive: https://embed.plnkr.co/Bebmjn2aAlflxYe1cwIS/

Similar to #421, but then new created sub menu's.

bbrala:
I understand that to get this to work nicely with Angular you would like to use promises like that. To be honest, I only have a small bit of experience with Angular. I can't really predict how much effort it would take to implement. 

I wouldn't be against a way to do this, but i wonder if it is even possible to do it properly without rewriting tons of code.

Anyways I won't be implementing this, but if someone with more experience wants to, it would be fine to talk about how we would do that.

Ruud-cb:
I @bbrala, Thanks for responding. Angular indeed covers this nicely, but is similar to [jquery promises](https://api.jquery.com/deferred.promise/) which seems to do the same. 
Thanks for marking it as a Feature, we need it any way so I might be implementing it. Of course, this would then need to be done without Angular, by using the Jquery promises. 
Indeed it would be a tough task to accomplish, I will first have to take a good look at the code, with some POC I will check if it is possible and then I will let you know for a possible pull request.

bbrala:
Would be quite awesome if you could find the time to do that! If you have any questions regarding the library, just put it here or send me an email. :)

Ruud-cb:
[...]

PR #431, opened 2016-07-22
Context menu async loading sub-menu's

Ruud-cb:
Hi,

Issue: #429
With this feature it is possible to load sub-menu items AFTER the context menu has been opened. By providing a [Jquery promise](http://api.jquery.com/deferred.promise/) for the "items". After the promise will be resolved or rejected the items will be shown in the context menu.

Example usage:

```javascript
    var errorItems = { "errorItem": { name: "Could not load items" },};//example usage if you want to reject promise
    var loadItems = function () {//example method that will eventually do an async call
        var dfd = jQuery.Deferred();
        setTimeout(function () {
            dfd.resolve(subItems);
        }, 1000);
        //setTimeout(function () {
        //    dfd.reject(errorItems);
        //}, 1000); //could be used to reject items, providing the same format list.
        return dfd.promise();//return a jquery promise, see http://api.jquery.com/deferred.promise/
    };

    var subItems = {
        "sub1": { name: "Submenu1", icon: "edit" },
        "sub2": { name: "Submenu2", icon: "cut" },
    };
    //normal context menu initialization.
    $.contextMenu({
        selector: '.context-menu-one',
        build: function ($trigger, e) {
            return {
                callback: function (key, options) {
                    var m = "clicked: " + key;
                    console.log(m);
                },
                items: {
                    "edit": { name: "Edit", icon: "edit" },
                    "cut": { name: "Cut", icon: "cut" },
                    "status": {
                        name: "Status",
                        icon: "delete",
                        items: loadItems(),//providing promise instead of items
                    },
                }
            };
        }
    });
````

working plunker:
https://embed.plnkr.co/iEkk5Ohrja86xf4UKObA/

**Implementation**
Looking at the code changes, which is commented quite extensively anyway: It checks on create for all sub-items if `item.items` is a promise. If so it will use another function `processPromises` within operations to deal with this. There, it will wait for either `promise.resolve` or `promise.reject`. If one of these occur it will override the `items` of the current node (`opt`) which will now contain actual items. Thereafter it will create the sub-menu.
We also call `op.update()` to put the `$node` on a correct x and y position, thereafter we can call `root.positionSubmenu()`, which will correctly position the sub-menu, only when the user already is hovering over the menu item.
Two more adjustments: inside `positionSubmenu` we have to check if $menu is undefined. Because `focusItem` calls this function anyway. This just prevents an console error.
And: When appending css styles when creating the context (sub) menu, add an extra check if it already has a 'sub' type. This because we call create again with an item that already has `sub` and should not execute the last else-if.

TODO:
- console error when closing context menu before promise is resolved
- Unit tests need to be created for this.
- optional nice to have: loading icon when promise is being resolved.

### DIFF
--- a/src/jquery.contextMenu.js
+++ b/src/jquery.contextMenu.js
@@ -1159,14 +1159,12 @@
                             }
                         });
                     } else {
-
                         // add label for input
                         if (item.type === 'cm_seperator') {
                             $t.addClass('context-menu-separator ' + root.classNames.notSelectable);
                         } else if (item.type === 'html') {
                             $t.addClass('context-menu-html ' + root.classNames.notSelectable);
-                        }
-                        else if (item.type === 'sub') {
+                        } else if (item.type === 'sub') {
                            //we don't want to execute the next else-if if it is a sub.
                         } else if (item.type) {
                             $label = $('<label></label>').appendTo($t);
@@ -1258,8 +1256,7 @@
                                 break;
                         }
 
-                        if (item.type === 'sub')
-                        {
+                        if (item.type === 'sub') {
                             //if item contains items, and this is a promise, we should create it later
                             //check if subitems is of type promise. If it is a promise we need to create it later, after promise has been resolved
                             if ('function' === typeof item.items.then) {
@@ -1269,7 +1266,6 @@
                                 // normal submenu.
                                 op.create(item, root);
                             }
-                           
                         }
                         // disable key listener in <input>
                         if (item.type && item.type !== 'sub' && item.type !== 'html' && item.type !== 'cm_seperator') {
--- a/src/jquery.contextMenu.js
+++ b/src/jquery.contextMenu.js
@@ -1,5 +1,5 @@
 /*!
- * jQuery contextMenu v@VERSION - Plugin for simple contextMenu handling
+ * jQuery contextMenu @VERSION - Plugin for simple contextMenu handling
  *
  * Version: @VERSION
  *
@@ -1061,7 +1061,6 @@
                 if (root === undefined) {
                     root = opt;
                 }
-
                 // create contextMenu
                 opt.$menu = $('<ul class="context-menu-list"></ul>').addClass(opt.className || '').data({
                     'contextMenu': opt,
--- a/src/jquery.contextMenu.js
+++ b/src/jquery.contextMenu.js
@@ -1232,11 +1232,18 @@
 
                             case 'sub':
                                 createNameNode(item).appendTo($t);
-
                                 item.appendTo = item.$node;
-                                //op.create(item, root); decide later, might be a promise.
                                 $t.data('contextMenu', item).addClass('context-menu-submenu');
                                 item.callback = null;
+								 //if item contains items, and this is a promise, we should create it later
+								//check if subitems is of type promise. If it is a promise we need to create it later, after promise has been resolved
+								if ('function' === typeof item.items.then) {
+									// probably a promise, process it, when completed it will create the sub menu's.
+									op.processPromises(item, root, item.items);
+								} else {
+									// normal submenu.
+									op.create(item, root);
+								}
                                 break;
 
                             case 'html':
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/jquery.contextMenu.js [hunk 1]
```diff
@@ -1227,7 +1234,7 @@
                                 createNameNode(item).appendTo($t);
 
                                 item.appendTo = item.$node;
-                                op.create(item, root);
+                                //op.create(item, root); decide later, might be a promise.
```
bbrala: Why isn't the logic for creating the submenu called here but evaluated later, outside the case statement?

src/jquery.contextMenu.js
```diff
@@ -1411,6 +1429,35 @@
...
+                    if (items === undefined) {
+                        //meh, null result, dev should have checked
+                        errorPromise(undefined);//own error object
+                    }
+                    finishPromiseProcess(opt,root, items);
+                };
+                function errorPromise(opt,root,errorItem) {
+                    //user called promise.reject() with an error item, if not, provide own error item.
+                    if ( typeof errorItem === 'string'  || errorItem === undefined) {
+                        errorItem = { "error": { name: "No items and no error item", icon: "context-menu-icon context-menu-icon-quit" } };
```
bbrala: Shouldn't this the errorItem as name if it is a string?

Ruud-cb: Hmm. I might had a reason for that but not sure why anymore, maybe to force users to build nice error items or something. I'll update it, one second.

bbrala:
I've started on cleanup and docs here.

https://github.com/swisnl/jQuery-contextMenu/tree/Ruud-cb-master

- [ ] Loading icon by using a class for a submenu waiting for a promise.
- [ ] Check if menu is still open after promise is resolved
- [ ] Add documentation
- [ ] Add tests.

Sweet-Bob:
@Ruud-cb Thank you! Very useful!

vherasme:
Hi,

I need to implement dynamic **submenu items** based on some properties of the content. Can I use this branch to do this ? Regards,

Victor

bbrala:
You could use this, but it is not finished yet, you could also use the `build` option which lets you build the moment you right click.

vherasme:
To be more specific. Can I use **build** to construct a list of items that will be of variable size ? Maybe I need to read more the documentation, but I am not sure I can do this. Could you please point me to an example ?

bbrala:
Sorry i was on mobile and getting a link was a bit hard.

https://swisnl.github.io/jQuery-contextMenu/demo/dynamic-create.html

vherasme:
I think this has a fixed number of subitems:

`                    "edit": {name: "Edit", icon: "edit"},
                    "cut": {name: "Cut", icon: "cut"},
                    "copy": {name: "Copy", icon: "copy"},
                    "paste": {name: "Paste", icon: "paste"},
                    "delete": {name: "Delete", icon: "delete"},
                    "sep1": "---------",
                    "quit": {name: "Quit", icon: function($element, key, item){ return 'context-menu-icon context-menu-icon-quit'; }}`

In my code, I will right click and depending on the properties I find, I will have to create an item for each of those properties. Also, that item's name will be the same as the found property. For example:

If i find, **dummy1, dummy2**, etc. I will create:

`"dummy1": {name: "dummy1"},
"dummy2": {name: "dummy2"},`

And of course, the names will always change as well. 

Thanks in advance for your patience. I know you take your time and effort to give this great plugin in for other developers. Cheers,

bbrala:
You could use the arguments from the build option (like $trigger) to figure out what menu items you want to show. The structure isn't set, your can return any menu you like in that function.

Sweet-Bob:
@bbrala Hooray! Thank you! Can you poke me in documentation, how i can use loading icon?

bbrala:
I think the docs are only updated on release. You could check demo/async-promise.md in the repository for a demo.

bbrala:
Documentation/demo/async-promise.md I mean

PR #445, opened 2016-08-29
Add options to support RTL and click to open sub-menu 

amnond:
To maintain compatability with swisnl/jQuery-contextMenu, ignore all files except src/jquery.contextMenu.js and src/sass/jquery.contextMenu-rtl.scss

options are:
rtl:true/false
push4sub:true/false


bbrala:
Thanks! I'll use this as a base for rtl support. :)

PR #451, opened 2016-09-05
make <input< and <select> tags xhtml compatible

andreasplesch:
- changed `<input>` to `<input />`
- removed second value attribute value=""; xhtml does not allow duplicated attributes (was there a reason for the second one?)
- added missing end tag `</select>` to `<select>` tag; also required by html


andreasplesch:
Thanks !

bbrala:
No problem, shouldn't result in any problems from what I gather we think.

PR #462, opened 2016-09-30
Add check for if opt.$menu is null when handling ESC key callback.

andreasrosdal:
Add check for if opt.$menu is null when handling ESC key callback.

This bug can be reproduced in the following way:
1. Go to Freeciv-web on https://play.freeciv.org and start a game.
2. Right-clicking on a unit, then the context menu is shown.
3. Press the ESC key.
4. Then a Javascript error occurs in the JavaScript console.

### DIFF
--- a/src/jquery.contextMenu.js
+++ b/src/jquery.contextMenu.js
@@ -639,7 +639,7 @@
 
                     case 27: // esc
                         handle.keyStop(e, opt);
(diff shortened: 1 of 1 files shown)


andreasrosdal:
I have applied this fix to Freeciv-web, so it can no longer be  reproduced there.

PR #467, opened 2016-10-24
Include src files on install

RoachMech:
In order to use the scss files we need to include the src directory on install


bbrala:
Hi @RoachMech 

Looked into it, and should be fine to do this. Guess we should change the `files` config to:

```
  "files": [
    "dist",
    "src"
  ],
```

Which should be fine.

RoachMech:
Thanks @bbrala for the quick merge

PR #469, opened 2016-10-25
Add check on destroy (Menu can be empty)

Sweet-Bob:
In my app, sometimes menu is empty (i.e. [o param is empty](https://github.com/swisnl/jQuery-contextMenu/blob/master/src/jquery.contextMenu.js#L1658)), i think this condition it will not be superfluous


bbrala:
Thanks, that is indeed a sane check.

PR #481, opened 2016-12-03
New Feature, selectable Sub Menus

zyuhel:
Adding ability to click on submenus. New option parameter selectableSubMenu, false by default


bbrala:
Thanks, that is usefull. And thank you for adding documentation :)

PR #487, opened 2016-12-12
All inputs had invisible letters.

EltonRst:
All inputs had invisible letters, making it impossible to read.

.context-menu-item.context-menu-hover {
  /*color: #fff; --  All inputs, textara and selects, font color #fff removed. */
  cursor: pointer; 
  background-color: #2980b9;
}


bbrala:
I've updates the sass to give inputs the default text color in a seperate commit. This is a commit in the dist instead of the sass :)

PR #498, opened 2017-01-12
moved in-line style for context-menu-layer

StealthDuck:
The inline style causes a Content Security Policy violation if _style-src 'unsafe-inline'_ is not defined in the policy.

Moved the style data to the CSS function below, which is permitted when a strict CSP is configured.


bbrala:
Thanks :)

PR #507, opened 2017-02-28
Two bugfixes

y-ich:
Hi.
Thank you for your nice plug-in!

I use your plug-in as delegated pattern with left click option, and found that  a function to disable
menu (by passing false as an argument) and a function to destroy menu (by passing'destroy' as an argument) do not work in the case of delegated pattern.

Here is a PR to fix them though I am not sure that it is a right way.

About the first issue, I fixed it in the case of left click option, but have no idea in defaul case (right click).

### DIFF
--- a/src/jquery.contextMenu.js
+++ b/src/jquery.contextMenu.js
@@ -353,9 +353,12 @@
             },
             // contextMenu left-click trigger
             click: function (e) {
+                var $this = $(this);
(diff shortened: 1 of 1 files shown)


bbrala:
Hi, thanks for taking the time to make a PR. Do you think you can make a fiddle with a demonstration of the problem so I can also test to see how things go with right click?

PR #511, opened 2017-03-13
Check for opt.$menu != null instead of opt.$menu !== null.

andreasrosdal:
This prevents calling trigger() on an undefined object. Based on my testing opt.$menu != null should be used to check for undefined, since this fixes the actual error based on my testing.

This is related to my previous pull request here:
https://github.com/swisnl/jQuery-contextMenu/pull/462

### DIFF
--- a/src/jquery.contextMenu.js
+++ b/src/jquery.contextMenu.js
@@ -507,7 +507,7 @@
                         });
                     }
 
-                    if (root !== null && root.$menu !== null) {
+                    if (root !== null && root.$menu != null) {
                         root.$menu.trigger('contextmenu:hide');
                     }
                 }, 50);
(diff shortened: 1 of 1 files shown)


bbrala:
Thanks for the PR!

When i changed this i wondered if that might break something, thanks for the report. I would prefer to check for null and undefined then though with `===`

andreasrosdal:
Sure, perhaps you can do that change? I'm mostly just interested in any solution which fixes this problem for the users.

bbrala:
Yeah, will do, btw, love your work :) Saw a post on civfanatics a bit back about your application. Love civ ^^

bbrala:
Updated and merged. :) Thanks again.

PR #513, opened 2017-03-17
check if given selected value is a 0, if it is a zero so return it as

Falseee:
Just check if the given selected Value is a boolean, if it is a int cast it to a string.

### DIFF
--- a/src/jquery.contextMenu.js
+++ b/src/jquery.contextMenu.js
@@ -1428,7 +1428,7 @@
                                 break;
 
                             case 'select':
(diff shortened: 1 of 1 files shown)


Falseee:
Just check if the given selected Value is a boolean, if it is a int cast it to a string.

Falseee:
Not working (before PR) : https://jsfiddle.net/s2yoqs8e/
Working (after PR): https://jsfiddle.net/s2yoqs8e/2/

PR #531, opened 2017-05-23
Fix #211 include triggering event in menu callback

### LINKED ISSUE #211: handle to triggering event in callback function
Hi Rodney, 
It will be nice to have the original event which triggered the context menu as part of the run-time options that are passed to callback function. I needed to know where the mouse of right clicked and I ended up doing this:-
        position: function($menu, x, y) {
                       //cache my x, y  
                       $.contextMenu.defaults.position($menu, x, y);
        },
Thanks,
sagar

wizzard0:
subj

### DIFF
--- a/src/jquery.contextMenu.js
+++ b/src/jquery.contextMenu.js
@@ -891,7 +891,7 @@
                 }
 
                 // hide menu if callback doesn't stop that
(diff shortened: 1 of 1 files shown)


bbrala:
Thanks for the contribution! Will have a look at it in the next week.

PR #550, opened 2017-07-07
left click

mtugnoli:
I it possible to show the menu with left mouse click ?


bbrala:
To awnser your question, yes ([docs for `trigger` option](https://swisnl.github.io/jQuery-contextMenu/docs.html#trigger)).

But seems you made a PR by accident?

PR #558, opened 2017-08-04
Add option to hide the menu on a second click/trigger of the trigger element

OliverColeman:
As the title says...
I was surprised to discover that the default behaviour when clicking on the trigger element a second time is to reposition the menu. However I had set-up the trigger to respond to left clicks, so given that right clicks are the default trigger action reposition makes more sense. 

It still seemed like a good idea to have the option of hiding the menu on a second trigger though; I think this would be the expected behaviour in the case of left-clicks or touch-screen taps (hence my initial surprise ;)). Perhaps this could even be the default behaviour for left-click or touch-screen taps? For now this PR is only adding the option and setting it to false by default.

### DIFF
--- a/src/jquery.contextMenu.js
+++ b/src/jquery.contextMenu.js
@@ -116,6 +116,10 @@
             // flag denoting if a second trigger should simply move (true) or rebuild (false) an open menu
             // as long as the trigger happened on one of the trigger-element's child nodes
             reposition: true,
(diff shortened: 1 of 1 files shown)


bbrala:
Thanks for the pr. Sounds like sane behaviour indeed. I'll have a look at it soon.

PR #565, opened 2017-09-03
Adding "activated" event and handling over boundary problem.

AliShahrivarian:
I added "activated" events so user can run desire function after contextMenu showed in page.

And a handler for over boundary problem with small resolution or even context menu with a lot of items.
In over boundary situation context menu will get scroll.


bbrala:
Thanks for the PR 👍 I'll have a look soon.

Issue #533, opened 2017-05-25
sub menu out of position


chitgoks:
please see attached. dev console in chrome is open to simulate this.

when the first context menu shows, it shows it above the dev console. once i navigate over to the sub menu, it does not position itself correctly.
![submenu](https://cloud.githubusercontent.com/assets/3955470/26473976/01465836-41e1-11e7-9cf9-405dd9d10873.png)

bbrala:
Do you use the `position.min.js` or just library itself?

chitgoks:
hi. only library itself

bbrala:
Ok thanks

bbrala:
This should be fixed in master right now. :)

See https://github.com/swisnl/jQuery-contextMenu/pull/565

chitgoks:
hi @bbrala it works now if jquery.position.min.js is included. so new fix requires this javascript file to be loaded?

AliShahrivarian:
hi @chitgoks , sorry, I forgot to check my PR on sub menu (because my issue was on root contextMenu). I will correct your issues in next days.

chitgoks:
hi ali that is ok i will wait

thank you for taking time for this

AliShahrivarian:
Hi again @chitgoks 
please use this two link and check if your problem solved:
[jquery.contextMenu.js](https://github.com/AliShahrivarian/jQuery-contextMenu/blob/master/src/jquery.contextMenu.js)
[jquery.contextMenu.css](https://github.com/AliShahrivarian/jQuery-contextMenu/blob/master/dist/jquery.contextMenu.css)

chitgoks:
![menu](https://user-images.githubusercontent.com/3955470/30255586-68a0b0da-96d7-11e7-80e8-71456da31feb.png)

Hi ali, problem not solved yet. this is the result

AliShahrivarian:
Ok, I got the problem.
I will fix it. (I'm going to work now :) I will check it in my free time)

bbrala:
Thanks for the effort @Alisharivarian

AliShahrivarian:
No worries @bbrala :+1: 
@chitgoks  could u check these files again? Sorry I'm using remote desktop and it's hard to test correctly.(it's too slow)
[jquery.contextMenu.js](https://github.com/AliShahrivarian/jQuery-contextMenu/blob/master/src/jquery.contextMenu.js)
[...]

PR #567, opened 2017-09-12
Fixed "sub menu out of position #533 "

AliShahrivarian:
Fixed sub menu overboundary and position problem.


bbrala:
As mentioned in #533 the `position: fixed` breaks the menu when scrolled down.

PR #696, opened 2019-09-13
Fast right click between two or more buttons Fix

feralonsomaccari:
If you click too fast beetwen two or more buttons, the layer on top of the everything stay bugged and force to reload the page. 
I created a new function and calling it solve the problem removing the layer 

To reproduce the bug you can try fast right clicking multiple times here and see the problem
https://swisnl.github.io/jQuery-contextMenu/demo/on-dom-element.html

### DIFF
--- a/src/jquery.contextMenu.js
+++ b/src/jquery.contextMenu.js
@@ -442,6 +442,9 @@
 
                 e.preventDefault();
 
+                // If you click it very fast beetwen two or more buttons the layer get bugged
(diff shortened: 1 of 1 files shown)


bbrala:
What browser is this, i cannot reproduce in firefox?

06Games:
I can confirm that this bug still happens. I use Firefox 86 and by right-clicking everywhere, the `context-menu-layer` div remains activated which then prevents any new interaction.

Issue #721, opened 2020-05-09
$layer insertBefore $menu cause page reflow


Rhain:
```js
layer: function (opt, zIndex) {
                // add transparent layer for click area
                // filter and background for Internet Explorer, Issue #23
                var $layer = opt.$layer = $('<div id="context-menu-layer"></div>')
                    .css({
                        height: $win.height(),
                        width: $win.width(),
                        display: 'block',
                        position: 'fixed',
                        'z-index': zIndex,
                        top: 0,
                        left: 0,
                        opacity: 0,
                        filter: 'alpha(opacity=0)',
                        'background-color': '#000'
                    })
                    .data('contextMenuRoot', opt)
                    .insertBefore(this)
                    .on('contextmenu', handle.abortevent)
                    .on('mousedown', handle.layerClick);

                // IE6 doesn't know position:fixed;
                if (typeof document.body.style.maxWidth === 'undefined') { // IE6 doesn't support maxWidth
                    $layer.css({
                        'position': 'absolute',
                        'height': $(document).height()
                    });
                }

                return $layer;
            },
```

the code `insertBefore(this)` cause document reflow. when the page is complex, it cause much time to reflow .
![image](https://user-images.githubusercontent.com/1488036/81467763-81e94500-920d-11ea-9060-f6d1209f3981.png)

Can you optimize this problem?

bbrala:
Merged :) Thanks!

PR #722, opened 2020-05-09
fix when add layer cause page forced reflow

Rhain:
see [#721](https://github.com/swisnl/jQuery-contextMenu/issues/721)


Rhain:
@bbrala  Can you pay attention to this problem。 thx

PR #728, opened 2020-08-12
Allow overriding default icon colors

RentecTravis:
Add `!default` after 
vars `$context-menu-icon-color` and `$context-menu-icon-color-hover`, so users of this module can supply their own values.


bbrala:
Ah cool, good point, thanks!

PR #767, opened 2022-10-31
added charset="UTF-8" and lang="en"

abhishektiwari47:
In the index.html I have added the language and the character set of the document.

### DIFF
--- a/test/index.html
+++ b/test/index.html
@@ -1,7 +1,8 @@
 <!DOCTYPE html>
-<html>
+<html lang="en" >
 <head>
     <title>QUnit Test Suite</title>
+    <meta charset="UTF-8">
     <link rel="stylesheet" href="../node_modules/qunitjs/qunit/qunit.css" type="text/css" media="screen">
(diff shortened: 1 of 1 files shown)


JaZo:
Hi @abhishektiwari47, can you explain why this is necessary for the tests?

abhishektiwari47:
May be it has no such changes in the project but it is always considered to have the DOCTYPE, language and the charset of the html document. It  increase the quality of the code and hence it is a good practice to do so.

abhishektiwari47:
Language 
It helps the search engines understand the correct language of the webpage. ...
Browser extensions and other software such as translation tools can use the lang tag to display the information in the correct language.

Charset
The HTML5 specification encourages web developers to use the UTF-8 character set, which covers almost all of the characters and symbols in the world!

JaZo:
I do know what these attributes/tags are for, but I don't see why this is needed for the unit tests. Besides, UTF-8 is already the default for HTML5 as it's [the only valid option](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#attr-charset). I'll close this PR, but feel free to reopen if you can explain why this is needed.

PR #768, opened 2023-01-10
Fix README.md inserting latex

Speedphoenix:


Speedphoenix:
I would also like to bump this other pull request that hasn't gotten any love in a while https://github.com/swisnl/jQuery-contextMenu/pull/760

PR #776, opened 2025-01-30
Support FontAwesome 6 classNames

RentecTravis:
Update FontAwesome-icon check to detect new FontAwesome icon styles introduced in v6

### DIFF
--- a/src/jquery.contextMenu.js
+++ b/src/jquery.contextMenu.js
@@ -1339,15 +1339,12 @@
                             if ($.isFunction(item.icon)) {
                                 item._icon = item.icon.call(this, this, $t, key, item);
                             } else {
-                                if (typeof(item.icon) === 'string' && (
-                                    item.icon.substring(0, 4) === 'fab '
-                                    || item.icon.substring(0, 4) === 'fas '
-                                    || item.icon.substring(0, 4) === 'fad '
-                                    || item.icon.substring(0, 4) === 'far '
-                                    || item.icon.substring(0, 4) === 'fal ')
+                                if (
+                                    typeof(item.icon) === 'string' 
+                                    && /^fa[lrsdbt]{1,3}\s+|fa-(?:solid|regular|light|thin|duotone|brands|sharp|sharp-duotone)\s+/.test(item.icon)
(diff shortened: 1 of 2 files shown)


bbrala:
Shouldn't we support both versions?

bbrala:
Thanks for this, and sorry it sat so long. Answering the "shouldn't we support both versions?" question: yes, and I've opened #821 which supports 4, 5, 6 and 7 together.

Two reasons I went with a fresh PR rather than building on this one.

**Enumerating styles is a losing game.** v7 keeps adding packs, each with its own short prefix. The docs list `sharp-duotone` (`fasds`) and `notdog-duo` (`fands`), and there is a [REST endpoint](https://docs.fontawesome.com/apis/rest) whose whole purpose is enumerating them, so any fixed list needs patching again at v8. #821 instead asks whether there is an `fa-*` token at all, and whether the class list already names a family/style. Short prefixes are matched by shape rather than from a list.

**Renaming `--fa5` to `--fa6` is a breaking change.** Anyone with custom CSS matching `.context-menu-icon--fa5` loses it on a minor upgrade, silently. The marker was never version-specific anyway, it is just the hook the stylesheet uses to position the child icon, so #821 keeps the name and documents it as historical.

Your diagnosis was right, and #821 fixes the case you identified plus three more that were also broken: `fa fa-user` (v4's full syntax, which only ever worked in the `fa-user` shorthand form), `fasds fa-user`, and `fa-brands fa-github`.

One thing worth flagging, since it is the trap here: `icon: "fa-user fa-lg"` works today *because* the plugin supplies `fa`. Passing classes through untouched whenever they look like v6 would have silently broken every v4 site using a modifier. #821 keeps supplying `fa` whenever the class list does not name a family itself, and pins that with a test.

Since this branch is also a few versions behind `master` now (`$.isFunction` has gone for jQuery 4 support, and the icon element is built with `addClass` rather than string interpolation since #810), I'd suggest closing this in favour of #821. Credited you in the changelog there.

RentecTravis:
Thanks for your work, and for providing the extensive reasoning behind your decisions.

PR #782, opened 2026-03-10
Replaced use of $.isFunction() for jQuery 4 compatibility.

AndrewDodd42:
jQuery 4 removes the $.isFunction() API and recommends replacing it with typeof checks.


bbrala:
Thanks, that should work cleanly :)

PR #784, opened 2026-07-27
fix: Fix test suite and add GitHub Actions CI

bbrala:
- Migrate acceptance tests (test/specs/*.js) to Playwright
- Switch gulp-sass to Dart Sass
- Use package-lock.json, drop yarn.lock.
- Remove Sauce Labs testing
- Add .github/workflows/tests.yml


bbrala:
Took a bit, but we have working tests again on all jQuery versions, also did some dev dependency updates to make sure we are up to speed again.

PR #787, opened 2026-07-28
fix: keep sub-menus visible when the parent menu becomes scrollable

### LINKED ISSUE #775: Overflow / visibility issue if contextmenu longer than screen height
If the context menu is too big for current screen size, these CSS values are added: `overflow-x: hidden` and `overflow-y: auto`, see:
https://github.com/swisnl/jQuery-contextMenu/blob/master/src/jquery.contextMenu.js#L1589

However, this causes that open sub-menus on the left/right of the current main menu aren't visible anymore. I've tried to change it to `overflow-x: visible`, but it still doesn't work, since it seems that `overflow-x: visible` is not possible in combination with `overflow-y: auto`.

bbrala:
Once a menu is too tall for the viewport it gets `overflow-y: auto`, but sub-menus are DOM descendants of that scrollable list, so they were being clipped along with it. `overflow-x: visible` can't fix this (per spec, it becomes `auto` too once `overflow-y` isn't `visible`).

Now, when a menu needs to scroll, its direct sub-menus are detached to `<body>` and positioned like a root menu, then moved back when the menu closes.

Fixes #775


bbrala:
Went through all 3 Codex findings, all confirmed real, all fixed.

**Detached submenus stayed visible after moving focus away (P1).** Nested submenus close automatically because `focusItem` clears the `visible` class from sibling `<li>`s unconditionally, and the CSS relies on that class via a `>` descendant selector. Detached submenus use a different CSS rule (a class on the `<ul>` itself, since it's no longer a DOM child of its opener), and that class was only ever cleared inside `blurItem`'s `if (opt.autoHide)` branch, default is `autoHide: false`, so with default settings the old submenu just stayed open and clickable. Now cleared unconditionally, matching how nested submenus already behave.

**Promise-supplied submenus never got detached (P2).** `detachSubmenus` runs once right after the parent menu is shown, and skips any submenu whose `$menu` doesn't exist yet. For a submenu built from a promise, that's exactly the case, it gets created later by `finishPromiseProcess`, which never re-ran the detach check. Fixed by re-checking the scroll state and detaching there too, plus fixing the positioning if the opener was already hovered before the promise resolved.

**No handling for scrolling the root menu while a detached submenu is open (P2).** There was no scroll handler at all, so a detached submenu (positioned once, in page coordinates) would drift out of alignment with its opener as the root list scrolled. Went with the simpler of the two options mentioned in the original comment: close the detached submenu on root scroll rather than trying to live-reposition it. Note this is a scoped fix, no live-follow, and keyboard selection state isn't restored after the close, that felt like reasonable scope for this PR rather than building a bigger tracking mechanism.

Added 3 new Playwright tests for these (verified each fails on the old code, passes on the new). Also merged in `master` since this branch had drifted behind the other PRs from this batch. Full suite: 27/27 unit, 15/15 Playwright.

PR #789, opened 2026-07-28
fix: upgrade embedded jQuery UI Position to 1.13.2 (CVE-2021-41184)

### LINKED ISSUE #765: Security - CVE-2021-41184: Upgrade embedded jQuery UI Position to 1.13+ (v2.x)
This plugin embeds jQuery UI Position v1.12 (https://github.com/swisnl/jQuery-contextMenu/blob/master/dist/jquery.ui.position.js).
This version of jQuery UI is susceptible to Cross Site Scripting and is officially unsupported by jQuery UI.

Upgrading this embedded code to jQuery UI 1.13 closes these known security vulnerabilities.

CVE: https://www.cve.org/CVERecord?id=CVE-2021-41184 / https://security.snyk.io/vuln/SNYK-JS-JQUERYUI-1767175
jQuery UI: https://jqueryui.com

bbrala:
The vendored `jquery.ui.position.js` was still 1.12.x, which has the `of`-option HTML-sniffing vulnerability fixed upstream in 1.13.0. Upgraded both the src and dist copies to the real 1.13.2 Position module.

This plugin always calls `.position()` with a DOM element as `of`, so the vulnerable path isn't reachable through normal usage here, this is a defense-in-depth fix.

Closes #765


bbrala:
Re: the Codex review comment about the removed `supportsOffsetFractions()` / IE 8-9 rounding fallback.

Checked this against real upstream: pulled `jquery-ui@1.13.2` fresh from npm and diffed its actual Position module against what's now vendored here, they're byte-identical. Upstream jQuery UI itself dropped that fallback in 1.13.x, it wasn't lost in how this file was assembled. So I didn't patch it back in, doing that would make this file diverge from the genuine, CVE-verified upstream source, which defeats the point of this PR.

The comment did surface a real (if minor) loose end though: `package.json` still declared `browserslist: "Explorer >= 8"`, which no longer matches reality now that the vendored code doesn't support IE 8/9 either way. Bumped that to `Explorer >= 11` and rebuilt the CSS (drops a few now-unneeded `-ms-` prefixes). Tests still pass (22/22).

PR #793, opened 2026-07-28
fix: left-click trigger leaked a synthetic contextmenu event to unrelated ancestor listeners

### LINKED ISSUE #754: bug?: defining left click as trigger type triggers "oncontextmenu" of trigger button
I have encountered an unexpected side-effect of the left-click handling in the code: we have a trigger button inside a slick-grid table structure, our slick-grid cells have a handler for `onContextMenu` and we are now facing a major issue with the trigger button! it's calling the slick-grid handler instead of calling the `contextMenu` creator. 

following GIF shows the issue. the yellow clicks are left-click, the red ones are right-click...

contextMenu-leftClick

reason for it is the code below. clicking the trigger button is calling the `contextMenu` event, which in return triggers the `onContextMenu` function defined later on in our slick-grid section.

any suggestions how to avoid this? 
* one solution is to detect that the `onContextMenu` was triggered artificially in the `slick-grid` code and not to continue that function. this would require that the code knows that the trigger (event target) is a `contextMenu` object. I have not been able to find any plugin command or helper which would facilitate this.
* another and well, a better solution is to rewrite the code below to avoid calling `contextMenu` event! technically, it should be possible to define 2 different sets of actions for left-clicking or right-clicking an element and this piece of code is forbidding it...

```js
//jquery.contextMenu.js:351 - v2.9.2
// contextMenu left-click trigger
click: function (e) {
  e.preventDefault();
  e.stopImmediatePropagation();
  $(this).trigger($.Event('contextmenu', {data: e.data, pageX: e.pageX, pageY: e.pageY}));
},
```

bbrala:
## Summary
Fixes #754. When a menu uses `trigger: 'left'` (also affects `'hover'`/`'touchstart'`), the plugin simulates a right-click by doing `$(this).trigger($.Event('contextmenu', ...))`. jQuery's `.trigger()` bubbles that synthetic event through every `contextmenu` handler bound on ancestor elements, not just the plugin's own delegated listener — so unrelated code elsewhere on the page (e.g. a grid component's own `contextmenu` handler) receives it and can't tell it apart from a real right-click.

Fix: in the three spots that synthesize a `contextmenu` event from a non-contextmenu native event (`handle.click`, `handle.mouseup`, `handle.mouseenter`), call the internal dispatcher directly (`handle.contextmenu.call(...)`) instead of triggering a bubbling event. Behavior and event data passed to the plugin are unchanged; it just no longer re-enters the DOM's bubbling chain.

## Test plan
- [x] Added a regression test binding a `contextmenu` handler on an unrelated ancestor and asserting it doesn't fire for a `trigger: 'left'` menu
- [x] `npm run test-unit` — 28/28 pass
- [x] `npm run test-accept` (Playwright) — 10/10 pass

### DIFF
--- a/src/jquery.contextMenu.js
+++ b/src/jquery.contextMenu.js
@@ -459,6 +459,16 @@
                 hoveract.timer = setTimeout(function () {
                     hoveract.timer = null;
                     $document.off('mousemove.contextMenuShow');
+
+                    // The trigger may have been removed from the document while
+                    // this delay was pending. Triggering a bubbling event used to
+                    // make that a silent no-op (a detached node can't bubble to
+                    // the delegated listener); calling the dispatcher directly
+                    // doesn't have that safety net, so check explicitly.
+                    if (!$.contains(document.documentElement, $this[0])) {
+                        return;
+                    }
+
                     $currentTrigger = $this;
                     // See handle.click for why we call the dispatcher directly
                     // instead of triggering a bubbling 'contextmenu' event.
--- a/src/jquery.contextMenu.js
+++ b/src/jquery.contextMenu.js
@@ -386,7 +386,19 @@
             click: function (e) {
                 e.preventDefault();
                 e.stopImmediatePropagation();
-                $(this).trigger($.Event('contextmenu', {data: e.data, pageX: e.pageX, pageY: e.pageY}));
+                // Invoke the dispatcher directly instead of triggering a real,
+                // bubbling 'contextmenu' event: a bubbling synthetic event is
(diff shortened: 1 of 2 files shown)


bbrala:
Good catch — addressed in bbf8f9a. The hover-trigger dispatch is delayed by \`e.data.delay\` (default 200ms); switching from a bubbling \`.trigger()\` to a direct dispatcher call meant the delayed callback no longer no-ops if the trigger element gets removed from the document during that window. Added a check that the trigger is still attached before dispatching, plus a regression test. Full suite still green (29/29 unit, 10/10 e2e).

PR #796, opened 2026-07-28
feat: allow selector to be an Element or jQuery object

### LINKED ISSUE #748: Allow the selector receive an Element or JQueryNode
Many developers creates elements programatically. Reference the selector by it's own path is not good in this case. So, the selector property could receive the element itself instead only receive the path string.

```javascript
const element = document.createElement('button');
el.innerHTML('Click Me')

$.contextMenu({
    // Allow the selector be an Element or JQueryNode Element
    selector: element // or $(element), to make it easy for the implementation with JQuery
})

document.body.appendChild(element)
```

bbrala:
Closes #748

## What

`$.contextMenu({ selector })` previously had to be a CSS selector string. This adds support for passing a raw `Element` or a jQuery object as `selector`, so context menus can be attached to elements created programmatically before they're inserted into the DOM (or otherwise not addressable by a stable selector).

```javascript
const element = document.createElement('button');
element.textContent = 'Click Me';

$.contextMenu({
    selector: element, // or $(element)
    items: {
        foo: { name: 'Foo', callback: function () { alert('Foo!'); } }
    }
});

document.body.appendChild(element);
```

## Why this needed more than "just pass it through"

`selector` is used pervasively as a *string*:
- `namespaces[o.selector] = o.ns` — used as an object key for `update()`/`destroy()` lookups.
- `.on('contextmenu' + o.ns, o.selector, o, handler)` — jQuery delegated binding requires a selector *string* for the delegation argument; jQuery doesn't support delegating via a raw Element/jQuery object.
- A reserved-classname guard (`o.selector.match(...)`) that assumes a string.

## Approach

- When `selector` is an Element or jQuery object, events are bound **directly** to the element(s) instead of delegating through the document/context (delegation-by-element isn't meaningful the way delegation-by-selector is). This also means the menu can be registered before the element is inserted into the DOM, matching the issue's example.
- Added a parallel `elementSelectors` registry (an array of `{el, ns}` pairs, matched by element identity) alongside the existing string-keyed `namespaces` map, so `destroy()` (both `$.contextMenu('destroy', element)` and the global `$.contextMenu('destroy')`) correctly tears down element-based registrations too.
- String selectors are completely unaffected — same delegated binding, same `namespaces` behavior as before.

## Scope / what's NOT covered

This is intentionally a minimal implementation covering the case from the issue:
- A single `Element`, or a jQuery object wrapping one or more elements — bound directly (non-delegated).
- Re-registering the *exact same* element without destroying the previous registration first will bind an additional, independent handler (same as what already happens today if you register the same string selector twice) — destroy first if you want a clean re-registration.

## Tests

- Added `test/unit/issue-748-element-selector.test.js` covering: a not-yet-attached `Element` as `selector`, a jQuery-wrapped element, destroy + re-registration on the same element (no throw / no double-bind), and that the global `destroy()` also tears down element-based registrations.
- `npm run test-unit` (karma + qunit, ChromeHeadless): 35/35 passing.
- `npm run test:fixtures && npx playwright test`: 15/15 passing, no regressions.

## Docs

Updated `README.md` and `documentation/docs/plugin-commands.md` to document the new supported `selector` types and the destroy-by-element usage.

Per project convention, `dist/` is rebuilt at publish time (`prepublishOnly`) and is intentionally not touched in this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,25 @@ $.contextMenu({
 });
 ```
 
+`selector` usually is a CSS selector string, delegated the same way as `$(document).on(event, selector, ...)`. It can also be a raw `Element` or a jQuery object, which is useful for elements that were created programmatically and aren't (yet) matched by a stable selector:
+
+```javascript
+var element = document.createElement('button');
+element.textContent = 'Click Me';
+
+$.contextMenu({
+    // bind directly to this element instead of delegating via a selector
+    selector: element, // or $(element)
+    items: {
+        foo: {name: "Foo", callback: function(key, opt){ alert("Foo!"); }}
+    }
+});
+
+document.body.appendChild(element);
+```
+
+Note: when `selector` is an `Element` or jQuery object, the contextMenu is bound directly to it rather than delegated, so registering with the exact same element again without destroying the previous registration first will bind a second, independent handler. Destroy it with `$.contextMenu('destroy', element)` (or `$.contextMenu('destroy', $(element))`) before re-registering.
+
 have a look at the [demos](http://swisnl.github.io/jQuery-contextMenu/demo.html).
 
 ## Version 3.0 beta
--- a/documentation/docs/plugin-commands.md
+++ b/documentation/docs/plugin-commands.md
@@ -59,7 +59,7 @@ To unregister / destroy a specific contextMenu:
 $.contextMenu( 'destroy', selector );
 ```
 
-selector expects the (string) selector that the contextMenu was registered to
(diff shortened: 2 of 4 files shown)


bbrala:
Addressed the Codex review comment (commit 59132ae): Element/jQuery-object selectors are now tracked in `elementSelectors` regardless of whether a custom `context` was supplied (previously guarded by `!_hasContext`, so `$(container).contextMenu({selector: element, ...})` left the direct binding on `element` untracked and therefore unremovable by any destroy path). The context-scoped destroy branch now also tears down any matching direct element bindings.

Added two regression tests covering this combination: a global `destroy()` and a `destroy(element)` call, both after registering via `$(container).contextMenu({selector: element, ...})`.

`npm run test-unit`: 37/37 passing. `npm run test:fixtures && npx playwright test`: 15/15 passing.

PR #797, opened 2026-07-28
feat: add direction option for RTL menu support

### LINKED ISSUE #742: Menu item text direction (or alignment)?
I was wondering how to customise the direction of the entire menu and its items, this is crucial for RTL apps.
If you haven't done it yet please add a feature for RTL support.

Thanks!

bbrala:
Closes #742

## Summary

Adds a `direction` option to `$.contextMenu({...})` (default `'ltr'`). When set to `'rtl'`:

- A `context-menu-rtl` class is added to the root menu **and** every sub-menu (including detached sub-menus that get moved to `<body>` when their parent menu is scrollable, see #775, so they don't lose the styling when they're no longer a DOM descendant of the root).
- CSS mirrors: text alignment (`direction: rtl; text-align: right`), icon placement (icon moves to the right edge instead of the left, covering both the icon-font and FontAwesome-5 code paths), checkbox/radio spacing, and the submenu-arrow triangle (now points left and sits on the left edge instead of the right).
- The submenu positioning logic (`positionSubmenu` in `src/jquery.contextMenu.js`) flips the jQuery UI `.position()` `my`/`at` arguments (and the equivalent fallback/detached-submenu math) so sub-menus open to the *left* of their parent item instead of the right, with the same `flipfit`/edge-clamping behavior as the existing LTR code so it still flips back to the right if there isn't enough room on the left.

## Scope notes

This covers the two things asked for in #742 (menu item text direction/alignment, and RTL support in general): a CSS hook for styling and a real flip of which side sub-menus open on. Root-menu placement itself already has viewport collision handling (`fit`) that isn't direction-specific, so it needed no changes.

## Files changed

- `src/jquery.contextMenu.js` – new `direction` default option, class applied at menu-creation time, `positionSubmenu` flip for both the jQuery UI code path and the plain-JS/detached-submenu fallback paths.
- `src/sass/jquery.contextMenu.scss` – `.context-menu-rtl` rules mirroring text alignment, icon placement, submenu arrow and initial submenu position.
- `documentation/docs.md` – documents the new `direction` option with an example.
- `test/unit/direction-rtl.test.js` – new QUnit regression tests: default menu has no rtl class, `direction: 'rtl'` adds the class to both the root menu and its sub-menus, and a sub-menu opens to the left of its parent item in `rtl` mode vs. to the right in the default `ltr` mode.

## Test plan

- [x] `npm run test-unit` (karma + qunit, ChromeHeadless) – 34/34 passing, including the 3 new RTL tests.
- [x] `npm run test:fixtures && npx playwright test` – 15/15 passing, no regressions.
- [x] Compiled the SCSS standalone (`npx sass`) to confirm it has no syntax errors.

`dist/` was intentionally left untouched per this repo's convention of rebuilding it at publish time.

### DIFF
--- a/documentation/demo/rtl-long-submenu.md
+++ b/documentation/demo/rtl-long-submenu.md
@@ -1,0 +1,55 @@
+---
+currentMenu: rtl-long-submenu
+---
+
+# Demo: RTL menu with a long (overflowing) sub-menu
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+
+- [Example code](#example-code)
+- [Example HTML](#example-html)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+<span class="context-menu-one btn btn-neutral">right click me</span>
+
+## Example code
+
+<script type="text/javascript" class="showcase">
+$(function(){
+    /**************************************************
+     * Combined regression coverage for `direction: 'rtl'`
+     * (#742) together with a sub-menu taller than the
+     * viewport (#752): the sub-menu must still open on
+     * the LEFT of its parent item (RTL) AND be detached,
+     * capped to the viewport and scrollable, exactly like
+     * the plain LTR case in long-submenu-short-root.md.
+     **************************************************/
+    var subItems = {};
+    for (var i = 1; i <= 40; i++) {
+        subItems['sub-key' + i] = {name: 'Sub item ' + i};
+    }
+
+    $.contextMenu({
+        selector: '.context-menu-one',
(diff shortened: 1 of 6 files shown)


bbrala:
Addressed the Codex review feedback in 58794b2:

- **Clamp flipped RTL submenus against the right viewport edge** (`src/jquery.contextMenu.js:270`): when a detached RTL sub-menu didn't fit to the left of its opener and flipped to the right, it wasn't checked against the right viewport edge afterwards (only the pre-existing left-edge clamp applied, which guards the wrong side once you've flipped rightward). Added a right-edge clamp for that flipped case, mirroring the left-edge clamp already used for the equivalent LTR flip-to-left case.

Verified with `npm run test-unit` (34/34) and `npx playwright test` (15/15) after the fix.

PR #798, opened 2026-07-28
fix: cap and scroll a sub-menu that's taller than the viewport

### LINKED ISSUE #752: Display bug for a long list of submenu
Hi,

I have a display problem for a submenu that contains more item that screen can to have and I can't scroll to see the rest... Can you help me please ?

Capture d’écran 2021-09-09 à 16 25 23

Best regards.

bbrala:
Closes #752

## Problem

#775 (merged in PR #787) fixed the case where the ROOT menu is taller than the viewport: it caps the root's height, makes it scrollable, and detaches its direct sub-menus to `<body>` so they aren't clipped by that scroll container.

It did not handle the reverse case, which is what #752 actually reports: a short root menu whose SUB-MENU has enough items to be taller than the viewport on its own. Nothing capped or scrolled that sub-menu, so items past the bottom edge of the screen were simply unreachable, no scrollbar, no room to flip.

## Fix

`positionSubmenu()` now detects when a sub-menu it's about to show is itself taller than the viewport. If so, it detaches the sub-menu to `<body>` (the same mechanism #775 already uses) and caps/scrolls it exactly like an overflowing root menu is handled in `op.activated()`, regardless of whether the root needed that treatment.

This sizing check only runs once per "show" of the sub-menu. `handle.focusItem()` calls `positionSubmenu()` again for every sibling item hovered inside an already-open sub-menu (they share the same opener/menu pair), so re-measuring and re-capping on every one of those calls would otherwise reset the sub-menu's scroll position back to zero mid-interaction - caught this while writing the regression test and fixed it by only doing the sizing work while the sub-menu isn't marked `context-menu-detached` yet; `op.reattachSubmenus()` clears the cap again before the next full reopen.

## Testing

- Added `test/specs/overflow-submenu-only.js` (Playwright): reproduces #752 with a short root menu + a 40-item sub-menu at a 1000x500 viewport, asserts the sub-menu stays within the viewport, is scrollable, and its last item is reachable/clickable. Also has a dedicated regression test for the scroll-reset bug described above.
- Added `documentation/demo/long-submenu-short-root.md` fixture used by the test.
- `npm run test:fixtures && npx playwright test` - all 17 tests pass (16 existing + 1 new suite with 2 tests), including the existing `overflow-submenu.js` (#775) suite, unaffected.
- `npm run test-unit` (karma/qunit) - all 31 tests pass.
- `npx eslint` - clean.

No changes to `dist/` (rebuilt at publish time per project convention).

### DIFF
--- a/documentation/demo/long-submenu-short-root.md
+++ b/documentation/demo/long-submenu-short-root.md
@@ -1,0 +1,53 @@
+---
+currentMenu: long-submenu-short-root
+---
+
+# Demo: Short menu with a long sub-menu (overflow)
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+
+- [Example code](#example-code)
+- [Example HTML](#example-html)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+<span class="context-menu-one btn btn-neutral">right click me</span>
+
+## Example code
+
+<script type="text/javascript" class="showcase">
+$(function(){
+    /**************************************************
+     * Context-Menu that itself fits the viewport, but
+     * whose sub-menu has many items (taller than the
+     * viewport) - reproduces #752, a variant of the
+     * overflow clipping issue (#775) where it's the
+     * sub-menu, not the root menu, that overflows.
+     **************************************************/
+    var subItems = {};
+    for (var i = 1; i <= 40; i++) {
+        subItems['sub-key' + i] = {name: 'Sub item ' + i};
+    }
+
+    $.contextMenu({
+        selector: '.context-menu-one',
+        callback: function(key, options) {
+            var m = "clicked: " + key;
+            window.console && console.log(m) || alert(m);
+        },
+        items: {
+            item1: {name: 'Item 1'},
+            item2: {name: 'Item 2'},
+            fold1: {
+                name: 'Long sub group',
+                items: subItems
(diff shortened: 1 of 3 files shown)


bbrala:
Fixed a real CI failure: the acceptance-test matrix was failing on jQuery 1.12.4 and 2.2.4 only, with the new sub-menu overflowing the viewport by ~0.17px in \`test/specs/overflow-submenu-only.js\`.

**Root cause:** jQuery's \`.outerHeight()\` getter (no argument) returns a fractional, \`getBoundingClientRect()\`-based value as of jQuery 3, but rounds to the nearest whole pixel on jQuery <3. \`positionSubmenu()\`'s \`maxTop\` clamp used that value alongside other still-fractional measurements (\`margin-top\`, \`$win.height()\`), so on old jQuery the clamp let the sub-menu sit slightly too low, and its real (unrounded) box then overshot the viewport bottom by the rounding error - consistently ~0.17-0.19px, matching the difference between the rounded and fractional height in this fixture.

**Fix:** added a small \`preciseOuterHeight()\` helper (plain \`getBoundingClientRect().height\`, bypassing jQuery's getter entirely) and used it for both the overflow check and the \`maxTop\` computation in \`positionSubmenu()\`. This is a real fix at the source, not a test-tolerance workaround.

Verified locally against all 4 pinned jQuery versions (1.12.4, 2.2.4, 3.7.1, 4.0.0) via \`JQUERY_VERSION=<version> npx playwright test\` - full suite (17 tests) passes on each, plus \`npm run test-unit\` (31/31) and eslint clean.

Issue #114, opened 2013-02-11
Issue with Select input command on Firefox


Annaleria:
I have a three tier nested contextMenu which works fine in Chrome, however in Firefox once I have opened the select input to make my selection, as soon as the mouse moves out of the input box the complete menu closes.  Options can be selected using the keyboard and everything then saves correctly.

I tried moving the select input to the second tier of the context menu and here it works perfectly.

I will try to set up a JSFiddle so I can test it further and you can fully see the problem.

---------------------------
http://jsfiddle.net/Annaleria/m4QTH/

I've set up a fiddle with the select input on both the second level and the third levels of the contextMenu.  You can see the difference in behaviour.  

In addition I have noted that on the second level, if the select options extend beyond the length of the menu then if the option clicked is beyond the menu then the menu will close upon the option being clicked, however if the option is over the menu then the menu will remain open after the option has been clicked.  

These behaviours can also be seen within the fiddle.  Please contact me if you require any further details or clarification.

PR #799, opened 2026-07-28
fix: don't treat a select's phantom mousedown as an outside click (#744)

### LINKED ISSUE #744: Bug option-select in submenu in firefox
On the third level of the menu (menu-submenu-submenu[select->option], when using the select, a bug appears when choosing an option. Firefox closes the submenu after selection. In EDGE everything is OK. My Firefox is ver. 84.0. Thanks.

bbrala:
Closes #744

## Problem

Choosing an option from a `type: 'select'` menu item nested inside a sub-sub-menu closed the whole menu tree in Firefox (Edge/Chrome were unaffected), as reported in #744.

## Root cause

`handle.layerClick` decides whether a `mousedown` is a genuine "click outside the menu" by looking at `document.elementFromPoint(x, y)` for the raw event coordinates and checking whether that lands inside `root.$menu`.

A native `<select>`'s options popup isn't part of the page's DOM/layout, so its rendered extent isn't constrained by the menu's own (possibly clipped) bounding box. Firefox is documented (see the plugin's own historical #114, where the original author diagnosed Firefox firing extra mouse events around a `<select>`'s native popup that don't correspond to a real page interaction) to fire an extra click/mousedown shortly after a selection, at coordinates that can fall outside that bounding box - even though the user's interaction never actually left the menu. `layerClick` then (correctly, by its own logic, but incorrectly for this case) treated that as an outside click and hid the menu.

## Fix

Track the timestamp of the last `change` event fired by any `type: 'select'` menu input, and briefly (500ms) suppress outside-click handling in `layerClick` right after it. A plain outside click with no preceding select interaction is unaffected; only a mousedown arriving shortly after choosing an option from a menu's own `<select>` is ignored.

## Tests

- `test/specs/select-in-submenu.js` (new):
  - Functional sanity check: choosing an option via `selectOption()` keeps the whole menu open (this alone would *not* have caught the bug - Playwright's `selectOption()` doesn't drive the browser's actual native popup, which no automation framework can do reliably across browsers/platforms - see comments in the test for detail).
  - A synthetic reproduction of the actual mechanism (`change` event + a `mousedown` dispatched directly on `#context-menu-layer` at coordinates below the menu's bounding box) that fails against the old code and passes with the fix.
  - A companion test confirming a genuine outside click with **no** preceding select change still closes the menu as before.
- `npm run test-unit` (karma/qunit, ChromeHeadless): 31/31 pass.
- `npm run test:fixtures && npx playwright test` (chromium): 18/18 pass.
- Manually verified locally against real Firefox 153 (installed via `npx playwright install firefox` + a temporary `firefox` project in `playwright.config.js`, not committed - see note below) with the same tests: all pass. I didn't keep the `firefox` project permanently since the acceptance-tests CI workflow (`.github/workflows/tests.yml`) only installs the `chromium` browser (`npx playwright install --with-deps chromium`); wiring up permanent Firefox CI coverage would need a workflow change too, which felt out of scope for this fix.

Only `src/`, `test/`, and `documentation/` are touched - no `dist/*` changes (rebuilt at publish time).

### DIFF
--- a/documentation/demo/sub-menu-select.md
+++ b/documentation/demo/sub-menu-select.md
@@ -1,0 +1,56 @@
+---
+currentMenu: sub-menu-select
+---
+
+# Demo: Select input nested in a sub-sub-menu
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+
+- [Example code](#example-code)
+- [Example HTML](#example-html)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+<span class="context-menu-one btn btn-neutral">right click me</span>
+
+## Example code
+
+<script type="text/javascript" class="showcase">
+$(function(){
+    /**************************************************
+     * Context-Menu with a `type: 'select'` item nested
+     * two levels deep, to reproduce #744 (choosing an
+     * option in Firefox closed the menu)
+     **************************************************/
+    $.contextMenu({
+        selector: '.context-menu-one',
+        callback: function(key, options) {
+            var m = "clicked: " + key;
+            window.console && console.log(m) || alert(m);
+        },
+        items: {
+            "fold1": {
+                "name": "Sub group",
+                "items": {
+                    "fold2": {
+                        "name": "Sub group 2",
+                        "items": {
+                            "my-select": {
+                                "name": "Choose",
+                                "type": "select",
+                                "options": {"opt1": "Option 1", "opt2": "Option 2", "opt3": "Option 3"},
+                                "selected": "opt1"
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    });
+});
+</script>
+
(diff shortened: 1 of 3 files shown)


bbrala:
Addressed the Codex review comment (thanks, valid catch): the previous commit's fix returned early - before the existing click pass-through and before the reposition/hideOnSecondTrigger logic - so a genuine outside click landing within the 500ms grace window would have been swallowed entirely rather than just failing to close the menu.

Pushed a follow-up (b4c019a) that narrows this:
- Only the final "hide the menu" step is withheld now; the pass-through to whatever's actually under the cursor always runs first, so a real outside click still reaches its target.
- It's also gated on position, not just time: only withheld when the click's coordinates plausibly fall within the just-changed `<select>`'s own horizontal span (where its native popup would render), not merely "any click within 500ms of any select changing anywhere on the menu."

Added a test (`a genuine outside click elsewhere shortly after a select change still closes the menu`) that fails against the previous commit and passes with this one, confirming the narrower guard resolves it without regressing legitimate outside-click behavior.

PR #800, opened 2026-07-29
fix: $.contextMenu('update') threw when a build menu had not been shown

### LINKED ISSUE #740: Disabling an entry gives " Cannot read property 'data' of null"
I'm using 3.0.0-beta.2 inside Chrome.
To update the enabled/disabled status of an entry of the menu the only solution I have found is to put a
```
disable: (): boolean => {
	return projectStatus === "ACTIVE";
}
```
in the entry and then have:
```
events: {
	show: (): boolean => {
		$.contextMenu("update");
		return true;
}
```
But when I open the menu the following error appears:
```
ContextMenuOperations.js:488 Uncaught TypeError: Cannot read property 'data' of null
    at HTMLUListElement.<anonymous> (ContextMenuOperations.js:488)
    at Function.each (jquery.min.js:2)
    at S.fn.init.each (jquery.min.js:2)
    at e.value (ContextMenuOperations.js:487)
    at e.value (ContextMenuOperations.js:521)
    at ContextMenu.js:278
    at Array.forEach (<anonymous>)
    at e.value (ContextMenu.js:277)
    at e.value (ContextMenu.js:53)
    at Function.c [as contextMenu] (contextmenu.js:34)
(anonymous) @ ContextMenuOperations.js:488
each @ jquery.min.js:2
each @ jquery.min.js:2
value @ ContextMenuOperations.js:487
value @ ContextMenuOperations.js:521
(anonymous) @ ContextMenu.js:278
value @ ContextMenu.js:277
value @ ContextMenu.js:53
c @ contextmenu.js:34
show @ prj-list.ts:143
value @ ContextMenuOperations.js:36
value @ ContextMenuEventHandler.js:100
dispatch @ jquery.min.js:2
v.handle @ jquery.min.js:2
```
Is this the correct way or there is another method to be used to enable/disable an entry (and change the name, but this is the next step).
Thanks for looking!
mario

bbrala:
## Problem

`$.contextMenu('update')` walks every registered menu and hands each one to `op.update()`, which immediately dereferences `opt.$menu`. A `build` menu only gets its `$menu` the first time it is actually shown, so as soon as a page has any `build` menu registered, a plain `$.contextMenu('update')` throws:

```
TypeError: Cannot read properties of undefined (reading '0')
```

The same happens for a registration that lingered as `null` after a failed `destroy`.

That is what the reporter of #740 ran into: they called `$.contextMenu('update')` from an `events.show` handler in order to refresh an item's `disabled` state, and the menu blew up on open. The stack in the issue is from the abandoned `3.x` branch, but `master` has the equivalent defect, reproduced here as a unit test.

While confirming this, a second problem in the same `case 'update':` block turned up: the context-scoped branch called `op.update($context)`, passing the context *element* where `op.update()` expects a menu's *options object*. It dereferences `$context.$menu`, which is always `undefined`, so `$.contextMenu('update', {context: ...})` could never do anything but throw.

## Fix

- `op.update()` returns early for a registration without a menu element, instead of throwing.
- The context-scoped branch now resolves the menus registered against that context and updates those.

## Note on the reporter's actual question

They asked whether calling `update()` from `events.show` is the right way to toggle an entry's enabled state per open. On 2.x it is not needed at all: `op.show()` already runs `op.update()` on every open, so a function-based `item.disabled` is re-evaluated each time the menu is shown. Added a sentence to the `disabled` docs saying so, plus a regression test covering it. The `update()` crash is still a real bug worth fixing, since `update()` is a documented public operation.

## Tests

`test/unit/issue-740-update.test.js`:

- `update()` from `events.show` does not throw and applies the `disabled` function
- `update()` does not throw when a `build` menu has never been shown (fails on master)
- `update()` from the `events.show` of a `build` menu does not throw (fails on master)
- `update()` scoped to a context updates that context's menu (fails on master)
- a `disabled` function is re-evaluated on every open without calling `update()`

`npm test` (47 tests) and the Playwright acceptance suite (22 tests) are green.

Closes #740

### DIFF
--- a/documentation/docs/items.md
+++ b/documentation/docs/items.md
@@ -178,6 +178,8 @@ Specifies if the command is disabled (`true`) or enabled (`false`).
 
 May be a callback returning a `boolean`. The callback is executed in the context of the triggering object (so this inside the function refers to the element the context menu was shown for). The first argument is the `key` of the command. The second argument is the `options object`.
 
+The callback is re-evaluated every time the menu is shown, so there is no need to call `$.contextMenu('update')` from an `events.show` handler to pick up a state change.
+
 `disabled`: `boolean` or `function(itemKey, opt)`
 
 #### Example
--- a/src/jquery.contextMenu.js
+++ b/src/jquery.contextMenu.js
@@ -1912,6 +1912,15 @@
             },
             update: function (opt, root) {
                 var $trigger = this;
+                // Nothing to update for a registration whose menu element doesn't
+                // exist (yet): a `build` menu only gets its $menu when it is first
+                // shown, and a destroyed registration can linger as null. Both are
+                // reachable from $.contextMenu('update'), which walks every
+                // registered menu, so bail out instead of throwing on $menu (see
+                // https://github.com/swisnl/jQuery-contextMenu/issues/740).
+                if (!opt || !opt.$menu || !opt.$menu.length) {
+                    return false;
+                }
                 if (typeof root === 'undefined') {
                     root = opt;
                     op.resize(opt.$menu);
@@ -2373,14 +2382,18 @@
 
             case 'update':
                 // Updates visibility and such
-                if(_hasContext){
-                    op.update($context);
-                } else {
-                    for(var menu in menus){
-                        if(Object.prototype.hasOwnProperty.call(menus, menu)){
-                            op.update(menus[menu]);
-                        }
+                for (var menu in menus) {
+                    if (!Object.prototype.hasOwnProperty.call(menus, menu)) {
+                        continue;
                     }
+                    // when a context was given, only update the menus registered
+                    // against it. Passing the context element itself to op.update()
+                    // (as this used to) can never work, it expects a menu's options
+                    // object and immediately dereferences its $menu.
+                    if (_hasContext && (!menus[menu] || menus[menu].context !== o.context)) {
+                        continue;
+                    }
+                    op.update(menus[menu]);
                 }
                 break;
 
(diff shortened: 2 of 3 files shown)


bbrala:
Thanks, both Codex points were legitimate and are addressed in 9420a73.

**"Update the active instance of build menus"** — correct, and the more important of the two. A `build` menu is rebuilt into a fresh options object on every invocation (`handle.contextmenu` reassigns `e.data`), and only that object ever gets a `$menu`. The registration kept in `menus` never does, so my guard turned the previous crash into a silent no-op rather than actually making `update()` work for build menus. Now the built options object is tracked per namespace in a `builtMenus` map, and the `update` operation refreshes that one when it exists. The entries need no explicit expiry, `op.hide()` already empties a build menu's options object when it hides, so the `$menu` guard picks it up from there. They are removed alongside their `menus` entry in every `destroy` path.

**"Bind scoped updates to the menu trigger"** — also correct, and it was not limited to the scoped path: the pre-existing `op.update(menus[menu])` in the non-context branch had the same problem, so function-based `disabled`, `visible`, `name` and `icon` callbacks have been receiving the internal `op` object as `this` from `$.contextMenu('update')` all along, instead of the trigger element the docs promise. Now invoked as `op.update.call(target.$trigger || $(), target)`. The empty-jQuery fallback keeps a `this.data(...)` / `this.hasClass(...)` in a user callback harmless for a registration that has never been shown.

Two regression tests added, both verified failing without the follow-up commit:
- `update() refreshes an open build menu, not just static ones`
- `update() runs function-based options against the trigger element` (asserts `this.jquery` and `this.is('.trigger')`, so it fails on the old `op`-as-`this` behaviour rather than only on the resulting class)

Nothing was deliberately left unaddressed. `npm test` (49) and the Playwright suite (22) are green locally.

bbrala:
👍

Issue #712, opened 2020-02-11
Bug in dataAttr?


sophanox:
I believe this bit of code:
```
                if(opt.dataAttr){
                    $.each(opt.dataAttr, function (key, item) {
                        opt.$menu.attr('data-' + opt.key, item);
                    });
                }
```

Should in fact be:

```
                if(opt.dataAttr){
                    $.each(opt.dataAttr, function (key, item) {
                        opt.$menu.attr('data-' + key, item);
                    });
                }

```
I.e., it should reference the key in the loop not the (non-existent) key property of opt? This at least makes it work for me

Cheers

bbrala:
You are right, thanks!

bbrala:
Fixed by #801, released in the next version.

The root cause was in `op.create()`: `dataAttr` was only read from the menu options and written as `'data-' + opt.key`, and `opt.key` is always undefined, so a menu with `dataAttr` ended up with a single `data-undefined` attribute holding the last value instead of one attribute per key. Item-level `dataAttr` was never applied at all, despite being documented since 2.9.0.

Both now go through a shared helper that iterates every key, converts camelCase to the dashed form jQuery's `.data()` getter expects (`{menuTitle: 'x'}` becomes `data-menu-title="x"`), and sets values via `.attr()` so they are never interpolated into markup.

PR #801, opened 2026-07-29
fix: apply dataAttr to menu items and use the right attribute names

### LINKED ISSUE #732: bug: Add `dataAttr` option for items
adding html 5 data- attr for each item is not working after checking the source code I found bug which is already reported here. #712 , but still if I fixed plugin not adding data attr to each menu item... 

so i did little changes to source code and add following snippets after line # 1183 // create contextMenu items

`if (item.dataAttr) {
 $t.attr('data-'+Object.keys(item.dataAttr), Object.values(item.dataAttr));
}`

by adding above snippets I can add data attr to menu items as following.

`var items = {
    firstCommand: {
        name: "Copy",
        dataAttr: {
            menuTitle: "My custom title"            
        }           
    }
}`

I hope this help someone ,, or maybe will be add in new release of plugin
thanks

bbrala:
## What

`dataAttr` was added in 2.9.0 and documented in `documentation/docs/items.md` as an **item** option, but the implementation only ever looked at `opt.dataAttr` (the menu options) and applied it to the menu's `<ul class="context-menu-list">`. Items never received any data attributes, which is what #732 reports.

The menu level path was broken too. It did:

```js
$.each(opt.dataAttr, function (key, item) {
    opt.$menu.attr('data-' + opt.key, item);
});
```

`opt.key` is always undefined, so instead of one attribute per key you got a single `data-undefined` attribute holding whichever value happened to come last. That is exactly what #712 describes.

## How

Both paths now go through one small helper, `applyDataAttr($element, dataAttr)`:

* iterates every key with `$.each`, so multi key objects work (the patch suggested in the issue only handled the first key)
* converts camelCase keys to the dashed form HTML uses, using the same rule jQuery's own `.data()` getter uses (`key.replace(/[A-Z]/g, '-$&').toLowerCase()`), so `menuTitle` is written as `data-menu-title` and reads back as `$item.data('menuTitle')` on jQuery 1, 2, 3 and 4. Keys that already use dashes are left alone.
* sets values with `.attr()`, so a value is never interpolated into an HTML string and cannot inject markup
* stringifies non string values (numbers, booleans) and skips `null` / `undefined` instead of writing `"null"` or removing attributes

Item attributes are applied on the `<li>` right after its `contextMenu` data is set, so they work for root items and sub-menu items alike. The menu level option keeps working, now with correct attribute names.

## Tests

New `test/unit/data-attr.test.js` covers a single key, multiple keys, camelCase to kebab conversion, the `.data()` round trip, non string and null/undefined values, values containing HTML not being interpreted as markup, sub-menu items, and the menu level case. Verified locally against jQuery 1, 2, 3 and 4, plus the Playwright acceptance suite.

Docs for the option have been expanded with the attribute naming rule, value handling, the rendered output and the menu level form.

Closes #732

Related to #712. The `data-undefined` bug reported there is fixed by the same helper, but I have left that issue open for a maintainer to close.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Unreleased
 
 * Context menu no longer jumps to the top of the screen #749
+* `dataAttr` is now applied to the menu items themselves, as documented, and writes every key of the object using the correct attribute name (fixes #732, fixes #712)
 
 ### 2.10.2
 
--- a/documentation/docs/items.md
+++ b/documentation/docs/items.md
@@ -24,6 +24,7 @@ currentMenu: items
   - [height](#height)
   - [items](#items)
   - [accesskey](#accesskey)
+  - [dataAttr](#dataattr)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
@@ -524,7 +525,13 @@ $.contextMenu({
 
 ### dataAttr
 
-Allows to pass data attributes (`data-*`) that get applied to the menu item. It should be passed as an object with key and a value.
+Allows to pass data attributes (`data-*`) that get applied to the menu item. It should be passed as an object of keys and values. Every key of the object is applied.
+
+Keys are converted from camelCase to the dashed form HTML uses, so `menuTitle` becomes `data-menu-title` and can be read back with `$item.data('menuTitle')`. Keys that already use dashes are left alone.
+
+Values are set as attributes, so they are never treated as HTML. Numbers and booleans are converted to their string form, `null` and `undefined` values are skipped.
+
+`dataAttr`: `object`
 
 #### Example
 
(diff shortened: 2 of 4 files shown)


bbrala:
Thanks, the Codex point about sub-menus was right and is fixed in a2d5958.

`op.create()` is called with the parent item as `opt` when it builds a sub-menu, so an item with both `items` and `dataAttr` had its attributes written twice: once on its own `<li>` and once on the sub-menu `<ul>`. The menu level application is now limited to the root menu (`opt === root`), which keeps the documented item level behaviour and leaves `$.contextMenu({dataAttr: ...})` working on the root `<ul>`. Added a regression test asserting the sub-menu `<ul>` does not carry the item's attribute and that only one element matches the attribute selector.

Nothing else from the review was skipped, that was the only suggestion.

bbrala:
👍

PR #804, opened 2026-07-29
fix: never evaluate caller-supplied selector strings as HTML

### LINKED ISSUE #731: Potentail XSS vulnerability
The input variable 'operation'
` // handle contextMenu triggers
    $.fn.contextMenu = function (operation) {
        var $t = this, $o = operation;
        if (this.length > 0) {  // this is not a build on demand menu
            if (typeof operation === 'undefined') {
                this.first().trigger('contextmenu');
   ...
   ...
            } else if ($.isPlainObject(operation)) {
                operation.context = this;
                $.contextMenu('create', operation);`

then call the function 'contexMenu', the operation pass to options, and pass to o 

`$.contextMenu = function (operation, options) {
        if (typeof operation !== 'string') {
            options = operation;
            operation = 'create';
        }

        if (typeof options === 'string') {
            options = {selector: options};
        } else if (typeof options === 'undefined') {
            options = {};
        }

        // merge with default options
        var o = $.extend(true, {}, defaults, options || {});`

in the end, o pass to contex，$-function evaluates its input as JavaScript-code instead of as a CSS-selector.The problem is that $(context) will execute JavaScript code if the value of context is a string like "".
` case 'destroy':
                var $visibleMenu;
                if (_hasContext) {
                    // get proper options
                    var context = o.context;
                    $.each(menus, function (ns, o) {

                        if (!o) {
                            return true;
                        }

                        // Is this menu equest to the context called from
                        if (!$(context).is(o.selector)) {
                            return true;
                        }`

bbrala:
Hardens selector handling so a caller-supplied string is never parsed as HTML.

## Problem

jQuery builds a detached DOM fragment from any string that looks like markup, rather than running it as a CSS selector. A value like `` passed where the plugin expects a selector therefore ends up building an element and running its inline handler instead of simply matching nothing.

Three options reached `$()` with a raw string:

- `context`, normalised once at the top of `$.contextMenu()`, so it applied equally to `create`, `update` and `destroy` (the `destroy` branch is the one reported in #731)
- `appendTo`, used in `op.create()`
- the `element` argument of `$.contextMenu.fromMenu()`

All three are documented as a selector string or a DOM element, never as markup.

## Change

A small `resolveSelector()` helper routes strings through `$(document).find()`, which only ever treats its argument as a selector. Anything that is not a string (Element, jQuery object, `document`) is handed to `$()` unchanged, so the existing supported inputs keep working, including the Element and jQuery object `selector` support added in #796.

A markup-looking string now either matches nothing or makes jQuery raise its usual "unrecognized expression" selector error, depending on whether it happens to be valid selector syntax. Neither builds an element.

## Tests

`test/unit/issue-731-selector-html-injection.test.js` covers each hardened call site with an `` payload whose `onerror` sets a global flag, asserting that the flag stays unset and that no `` is added to the document. It also covers the inputs that must keep working: `context` as a selector string, as an Element and as a jQuery object, and `appendTo` as a selector string and as an Element.

Verified failing before the change and passing after, on jQuery 1, 2, 3 and 4. Acceptance tests and eslint are unchanged.

## Not included

- `item.icon` is interpolated into markup for the Font Awesome paths (`$('<i class="' + item.icon + '"></i>')`). That is a different mechanism and `icon` may well be considered trusted config, so it is left alone here. `$('<i></i>').addClass('fa ' + item.icon)` would be an equivalent, non-parsing alternative if you want it changed.
- `inputLabel()` builds `label[for="' + node.id + '"]`. The id comes from the DOM and the string never starts with `<`, so it is not parsed as HTML, but a quote in an id would still break the selector.
- `$.contextMenu('update', {context: ...})` throws for unrelated, pre-existing reasons (`op.update()` receives the context itself, which has no `$menu`). Left as is; the test for that path only asserts the string is not evaluated.

Closes #731

### DIFF
--- a/src/jquery.contextMenu.js
+++ b/src/jquery.contextMenu.js
@@ -1772,7 +1772,7 @@
                 if (!opt.$node) {
                     opt.$menu.css('display', 'none').addClass('context-menu-root');
                 }
-                opt.$menu.appendTo(opt.appendTo || document.body);
+                opt.$menu.appendTo(resolveSelector(opt.appendTo || document.body));
             },
             resize: function ($menu, nested) {
                 var domMenu;
@@ -2265,6 +2265,23 @@
             (selector.nodeType === 1 || (typeof selector.jquery !== 'undefined' && typeof selector.length === 'number'));
     }
 
+    // resolve a caller-supplied "selector-ish" option (`context`, `appendTo`,
+    // the element passed to `fromMenu()`, ...) to a jQuery object without ever
+    // letting a string be evaluated as HTML. `$(string)` builds a detached DOM
+    // fragment whenever the string looks like markup instead of running it as a
+    // CSS selector, which turns something like `<img src=x onerror=...>` into
+    // executing code. `.find()` only ever accepts a selector, so strings are
+    // routed through it. Elements, jQuery objects, `document` and everything
+    // else are handed to `$()` unchanged.
+    // See https://github.com/swisnl/jQuery-contextMenu/issues/731
+    function resolveSelector(target) {
+        if (typeof target === 'string') {
+            return $(document).find(target);
+        }
+
+        return $(target);
+    }
+
     // remove every `elementSelectors` entry (and its bound handler) registered
     // under the given namespace. Used to tear down direct element/jQuery-object
     // bindings from any destroy code path, regardless of whether the menu was
@@ -2364,7 +2381,7 @@
             o.context = document;
         } else {
             // you never know what they throw at you...
-            $context = $(o.context).first();
+            $context = resolveSelector(o.context).first();
             o.context = $context.get(0);
             _hasContext = !$(o.context).is(document);
         }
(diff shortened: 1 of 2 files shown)


bbrala:
Thanks, addressed in ebba1e5.

**Addressed:** the destroy assertions in the "supported selector inputs" tests could not fail. The diagnosis was right, the predicted outcome was not: the tests passed on all nine CI jobs, because the menu was already open at that point and a second trigger on an open menu repositions rather than firing `show` again, so `shown` stayed at 1 whether or not `destroy` did anything.

The underlying point stands, so those tests were rewritten:

- the `context` tests now prove scoping. A trigger inside the context opens the menu, an identical trigger outside it does not. Both would fail if the string were resolved to a detached node.
- the `destroy` tests no longer open the menu first, and use a context that resolves to an element matching the menu's selector, which is the case `destroy` actually handles. Verified they fail when the context is pointed at a non-matching selector.

**Not addressed, reported instead:** while rewriting these, a raw `Element` passed as `context` turned out not to scope anything either. An Element has no `length`, so `!o.context.length` sends it down the "no context" branch and it silently becomes `document`. Only jQuery objects and selector strings scope a registration today. That is pre-existing and unrelated to this change, so it is left alone here rather than folded into a security fix. The test documents the current behaviour so a later fix has something to update.

bbrala:
👍

PR #807, opened 2026-07-29
feat: separate show/hide animation durations and optional silent re-open

### LINKED ISSUE #739: Animation Enhancements
1. An animation will replay when the context menu is visible and another right-click occurs. There should be a way to disable animations if the context menu is already visible.

2. There should be a way to assign a separate duration for show and hide animations.

Thank you.

bbrala:
Closes #739

Both requests from the issue, implemented backwards compatibly.

## 1. Don't replay the animation when the menu is already visible

New option `animation.animateOnReopen`, default `true` so nothing changes for existing configs. With it set to `false`, opening a menu that is already on screen (right clicking another element that shares the same menu) moves it to the new position instead of playing the hide and show animation again.

Two things were needed for that:

* `op.show()` samples whether `opt.$menu` itself is visible before anything is hidden, because hiding only *starts* an animation. This deliberately looks at the menu element, not at "is any menu open", so a different menu opening while another one closes is still animated. When the menu is being re-opened and re-animating is disabled, any hide animation is jumped to its end (its completion callbacks still run) and the menu is put back into its shown state synchronously, so it is never painted as hidden.
* With the default `useModal: true`, `handle.layerClick()` waits for `contextmenu:hidden` before opening the menu on the new trigger. That wait is what makes the menu collapse and expand again, so it is skipped for this case and the menu is opened right after the hide is triggered. Menus built on invocation (`build`) still wait, since they get a brand new menu element every time and are torn down in the hide callback.

## 2. Separate show and hide durations

New options `animation.showDuration` and `animation.hideDuration`, both `null` by default and falling back to `animation.duration`, which keeps working exactly as before as a single value for both directions. `0` is honoured as an override rather than treated as unset.

## Tests

* `test/unit/issue-739-animation.test.js`: default durations unchanged (asserted against the real `slideDown`/`slideUp` calls), `duration` still applying to both directions, per direction overrides, fallback behaviour, re-open still animating by default, and no replay with `animateOnReopen: false` (including the menu never collapsing). Green on jQuery 1, 2, 3 and 4.
* `test/specs/animation-reopen.js`: real right click on a second trigger through the modal layer, sampling the menu while it is re-opened and asserting it never fades out or disappears. Green on jQuery 1.12.4, 2.2.4, 3.7.1 and 4.0.0. It fails on `master`.

## Docs

`animation` in `documentation/docs.md` now documents all keys in a table, plus a new demo page `documentation/demo/animation.md` (two triggers sharing one menu, fade in/out with different durations and `animateOnReopen: false`). The documented default duration was also corrected from 500 to 50, which is what the source has always used.

### DIFF
--- a/documentation/_data/nav.js
+++ b/documentation/_data/nav.js
@@ -33,6 +33,7 @@ module.exports = [
       { id: 'simple-context-menu', text: 'Simple Context Menu', url: '/demo.html' },
       { id: 'fontawesome-icons', text: 'FontAwesome icons', url: '/demo/fontawesome-icons.html' },
       { id: 'accesskeys', text: 'Accesskeys', url: '/demo/accesskeys.html' },
+      { id: 'animation', text: 'Animation Options', url: '/demo/animation.html' },
       { id: 'async-create', text: 'Create Context Menu (asynchronous)', url: '/demo/async-create.html' },
       { id: 'async-promise', text: 'Create Context Menu (promise)', url: '/demo/async-promise.html' },
       { id: 'callback', text: "Command's action (callbacks)", url: '/demo/callback.html' },
--- a/documentation/demo.md
+++ b/documentation/demo.md
@@ -69,6 +69,7 @@ title: jQuery contextMenu — Demo gallery
 *   [Disabled Callback Command](demo/disabled-callback.html)
 *   [Changing Command's disabled status](demo/disabled-changing.html)
 *   [Accesskeys](demo/accesskeys.html)
+*   [Animation Options](demo/animation.html)
 *   [Submenus](demo/sub-menus.html)
 *   [Input Commands](demo/input.html)
 *   [Custom Command Types](demo/custom-command.html)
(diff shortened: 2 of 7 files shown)


bbrala:
Thanks, addressed the Codex review.

**Verify the target belongs to this menu before skipping the wait (P2)** - valid, fixed in a376fb1. `reopensSameMenuWithoutAnimation()` now also requires the trigger to sit inside this registration's `context`, so a trigger matching the same selector in another registration (another container with its own menu) no longer counts as a re-open and keeps waiting for the hide to finish as before. Matching on the selector alone was too loose there.

Not changed, for the record: the check stays selector plus context based rather than resolving the target's actual menu object. The plugin has no lookup from an arbitrary element to its registration (`namespaces` only holds document-context registrations), and selector plus context is exactly what the delegated `contextmenu` binding uses to decide which menu a trigger gets, so the two agree. Registering the very same selector twice on the same context is already ambiguous in the plugin itself, independent of this change.

bbrala:
👍

PR #808, opened 2026-07-29
fix: autoHide never applied to a nested trigger with a different trigger mode

### LINKED ISSUE #727: Auto hide does not work for children elements
My case: I want the user to right-click on a row in the table or left-click on a button in this row.
In this configuration, the `autoHide` for the internal element does not work. Perhaps this can be fixed somehow?
Example:
```
<span class="context-menu-two">Click me right! 
       <span class="context-menu-one">Click me!</span>
</span>
```
JavaScript:
```
           $(function() {
                $.contextMenu({
                    selector: '.context-menu-one', 
                    autoHide: true,
                    trigger: 'left',
                    callback: function(key, options) {
                        var m = "clicked: " + key;
                        window.console && console.log(m) || alert(m); 
                    },
                    items: {
                        "edit": {name: "Edit", icon: "edit"},
                        "cut": {name: "Cut", icon: "cut"},
                    copy: {name: "Copy", icon: "copy"},
                        "paste": {name: "Paste", icon: "paste"},
                        "delete": {name: "Delete", icon: "delete"},
                        "sep1": "---------",
                        "quit": {name: "Quit", icon: function(){
                            return 'context-menu-icon context-menu-icon-quit';
                        }}
                    }
                });


            });


            $(function() {
                $.contextMenu({
                    selector: '.context-menu-two', 
                    autoHide: true,
                    trigger: 'right',
                    callback: function(key, options) {
                        var m = "clicked: " + key;
                        window.console && console.log(m) || alert(m); 
                    },
                    items: {
                        "edit": {name: "Edit", icon: "edit"},
                        "cut": {name: "Cut", icon: "cut"},
                    copy: {name: "Copy", icon: "copy"},
                        "paste": {name: "Paste", icon: "paste"},
                        "delete": {name: "Delete", icon: "delete"},
                        "sep1": "---------",
                        "quit": {name: "Quit", icon: function(){
                            return 'context-menu-icon context-menu-icon-quit';
                        }}
                    }
                });


            });

```

bbrala:
## Problem

Registering two menus on nested elements with different trigger modes, the setup from #727, a `trigger: 'left'` button inside a row registered with `trigger: 'right'`, both with `autoHide: true`, leaves the inner menu unusable once the outer one has been opened, and autoHide never kicks in for it.

## Root cause

Three things go wrong, all in the hand-off between the menu that is closing and the one that should open.

1. `handle.layerClick()` decided whether the click that dismissed the open menu should also open another one from the **closing** menu's own `trigger` option (`triggerAction`). A left-click while a right-click menu is open therefore never re-dispatched, so the nested trigger's menu simply never showed. The button is now also matched against the triggers actually registered for the element under the cursor. The menu that gets asked to show still validates the button against its own `trigger` option in `handle.contextmenu()`, so nothing new can slip through.
2. Resolving what the click landed on required `root.$layer` to still be present 50ms later. autoHide's own timer routinely closes the menu inside that window, and the layer element lingers in the document for another few milliseconds after `root.$layer` is gone, so `elementFromPoint()` reported the layer itself as the click target.
3. `op.hide()` unbound the document-level keyboard and autoHide handlers unconditionally. A menu shown while the previous one was still tearing down lost its handlers again immediately and never auto hid. The same applied to `$currentTrigger`. Both are now only cleared when they still belong to the menu being hidden.

## Tests

New demo page `documentation/demo/nested-triggers-autohide.md` with the reporter's configuration, plus `test/specs/nested-triggers-autohide.js` covering:

- the inner (left-click) menu auto hides on its own
- the outer (right-click) menu auto hides on its own
- the inner menu opens, replaces the outer one, and auto hides while the outer menu is open
- the outer menu opens, replaces the inner one, and auto hides while the inner menu is open

The last two fail on `master` and pass with this change. The single-trigger cases pass either way and are kept as regression cover. Both flows also assert that no `#context-menu-layer` is left behind, so no orphaned menu stays on screen.

Unit tests, acceptance tests against jQuery 1.12.4 / 2.2.4 / 3.7.1 / 4.0.0, eslint and `npm run docs:build` are all green.

Closes #727

### DIFF
--- a/documentation/_data/nav.js
+++ b/documentation/_data/nav.js
@@ -54,6 +54,7 @@ module.exports = [
       { id: 'menu-title-fontawesome', text: 'Menu title with Font Awesome icons', url: '/demo/menu-title-fontawesome.html' },
       { id: 'menu-promise', text: 'Menu with promise', url: '/demo/menu-promise.html' },
       { id: 'on-dom-element', text: 'Context Menu on DOM Element', url: '/demo/on-dom-element.html' },
+      { id: 'nested-triggers-autohide', text: 'Nested Triggers With Autohide', url: '/demo/nested-triggers-autohide.html' },
       { id: 'sub-menus', text: 'Submenus', url: '/demo/sub-menus.html' },
       { id: 'sub-menus-promise', text: 'Submenus (promise)', url: '/demo/sub-menus-promise.html' },
       { id: 'trigger-custom', text: 'Custom Activated Context Menu', url: '/demo/trigger-custom.html' },
--- a/documentation/demo.md
+++ b/documentation/demo.md
@@ -63,6 +63,7 @@ title: jQuery contextMenu — Demo gallery
 *   [Swipe Trigger](demo/trigger-swipe.html)
 *   [Hover Activated Context Menu](demo/trigger-hover.html)
 *   [Hover Activated Context Menu With Autohide](demo/trigger-hover-autohide.html)
+*   [Nested Triggers With Autohide](demo/nested-triggers-autohide.html)
 *   [Custom Activated Context Menu](demo/trigger-custom.html)
 *   [Disabled Menu](demo/disabled-menu.html)
 *   [Disabled Command](demo/disabled.html)
--- a/documentation/demo/nested-triggers-autohide.md
+++ b/documentation/demo/nested-triggers-autohide.md
@@ -1,0 +1,65 @@
+---
+currentMenu: nested-triggers-autohide  
+---
+
+# Demo: Nested Triggers With Autohide
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+
+- [Example code](#example-code)
+- [Example HTML](#example-html)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+<span class="context-menu-two btn btn-neutral" style="display: block; width: 480px; height: 240px;">
+    <span class="context-menu-one btn btn-neutral">left click me!</span>
+    <br>
+    right click anywhere in this box
+</span>
+
+<div style="height: 240px;"></div>
+
+## Example code
+
+<script type="text/javascript" class="showcase">
+$(function(){
+    var items = {
+        "edit": {name: "Edit", icon: "edit"},
+        "cut": {name: "Cut", icon: "cut"},
+        "copy": {name: "Copy", icon: "copy"},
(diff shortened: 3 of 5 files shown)


bbrala:
Thanks, that one is fair and now addressed in 8b66cc2.

**Addressed:** the P2 about bubbling a synthetic `contextmenu` for nested left-clicks. Going through `$.fn.contextMenu`'s coordinate overload did reintroduce exactly the leak #754 removed from `handle.click`: a jQuery `contextmenu` listener bound on an ancestor of the nested trigger fired on a plain left-click. The trigger lookup now returns the matched element together with its options, so the menu is dispatched straight to `handle.contextmenu` instead. Two details came with that:

- it resolves the **innermost** match, mirroring which delegated handler a bubbling event would have reached first, so nested registrations with the same trigger mode still pick the closest one;
- it skips selector-string registrations whose `context` does not contain the element, since those are delegated from that context and would otherwise be over-matched by a direct dispatch. Element/jQuery-object selectors are bound to the element directly and are deliberately not scoped that way.

The new assertion in `test/specs/nested-triggers-autohide.js` fails against the previous commit and passes now, so the leak is covered going forward.

**Not changed:** the `$(target).contextMenu({x, y, button})` call is kept as the fallback for the case where only `triggerAction` matched and the lookup found nothing. That path is unchanged from `master`, so this PR does not widen it, and reworking it is out of scope here.

Unit tests, acceptance tests against jQuery 1.12.4 / 2.2.4 / 3.7.1 / 4.0.0 and eslint are green locally.

bbrala:
Rebased onto `feat/issue-739-animation-enhancements` (#807) and retargeted this PR's base to that branch, so the diff here stays limited to the #727 work. GitHub will retarget it back to `master` automatically once #807 merges.

The rebase had one real conflict. Both PRs restructure the same deferred re-open block in `handle.layerClick`, for different reasons:

- #807 wraps the re-open so it can skip the wait when the same menu is re-opened with `animation.animateOnReopen: false`, and fires it *after* `contextmenu:hide` so `op.show()` still samples the menu as visible.
- This PR replaces the block with `matchTriggerForClick` plus direct dispatch, and in its immediate case fired *before* the hide.

Taking either side wholesale drops the other's fix, so the two conditions are now combined: wait for `contextmenu:hidden` only when the menu is still active *and* this is not a no-animation re-open. The immediate case assigns `openTargetMenu`, which is invoked just after the hide block rather than inside it, so it still runs when the menu was already gone (the autoHide race this PR is about).

Verified together, not just separately: `animation-reopen.js` and all four `nested-triggers-autohide.js` specs pass in the same run, alongside the full unit suite.

bbrala:
👍

Issue #726, opened 2020-07-01
Using this library on Google Maps


glaukon-ariston:
I would like to be able to use this library on Google Maps. The Google Maps provides `google.maps.event.addListener` API to register `rightclick` event and then the callback receives and `event` object containing `pixel` (with `(x,y)` coordinates), `latLng` and possibly `feature` object if there is a geo feature under the mouse. I would like to use this data to provide context sensitive menu e.g. if right clicked on a geo feature object to list actions available on that specific object. 

Any idea if that's possible out of the box without changing the library?

Here's an old [Gmaps-Context-Menu](https://github.com/gizzmo/Gmaps-Context-Menu/blob/master/code.js) library for Google Maps as an example of how it works.

PR #816, opened 2026-07-29
fix: detect the {x, y} overload by key presence (#812)

### LINKED ISSUE #812: $.fn.contextMenu({x, y}) with undefined coordinates fails with a confusing "No selector specified"
Found while writing acceptance tests for #735.

## Problem

`$.fn.contextMenu(operation)` in `src/jquery.contextMenu.js` checks for the coordinate overload before the plain-object branch. When `x` and `y` are present as keys but `undefined` in value, the coordinate branch does not match, so the object falls through to the `$.isPlainObject()` branch and is treated as a **menu definition**. That reaches `$.contextMenu('create', ...)`, which throws:

```
No selector specified
```

That error is very confusing for what the caller clearly intended as a positioning argument. It is easy to hit when the coordinates come from an event that may not carry them:

```js
// e.pageX / e.pageY are undefined for a synthetic or keyboard-originated event
$(el).contextMenu({x: e.pageX, y: e.pageY});
// -> "No selector specified"
```

## Suggested fix

Either treat an operation object that has `x`/`y` keys at all as the coordinate overload and validate the values, giving a clear error such as "contextMenu: x and y must be numbers", or fall back to the element-relative default position when they are `undefined`.

## Related

While confirming this, the coordinate space of that overload turned out to be undocumented too. `{x, y}` are **page** coordinates, the same space as `event.pageX`/`event.pageY`, scroll included. They are not viewport coordinates and not relative to the trigger. `documentation/docs/plugin-commands.md` shows the call under "Manually show a contextMenu" without saying which space it expects, which is the first thing anyone integrating an external event source (a map, a canvas, a charting library) needs to know. See #726 for a worked example.

bbrala:
Closes #812

## The bug

`$.fn.contextMenu(operation)` decided between the `{x, y}` positioning overload and a menu definition with:

```js
} else if (typeof operation.x !== 'undefined' && typeof operation.y !== 'undefined') {
```

An event that carries no pointer position (a keyboard-originated or synthetic one) has `pageX`/`pageY` undefined, so `$(el).contextMenu({x: e.pageX, y: e.pageY})` produced `{x: undefined, y: undefined}`. That failed the test above, fell through to the `$.isPlainObject()` branch, was treated as a menu definition and reached `$.contextMenu('create', ...)`, which threw `No selector specified` for what was clearly a positioning call.

## The fix

Detection is now by key presence rather than by value: an object carrying an `x` or a `y` key is a positioning call regardless of what those values are. A menu definition still wins whenever the object also carries a definition-only key (`items`, `selector` or `build`), so an options object that happens to have an `x`/`y` of its own cannot be misrouted. The plugin has no `x`/`y` *option*, so that guard is belt and braces, but it keeps the widened detection strictly non-breaking.

For the values themselves I chose the **graceful fallback**, not a thrown error: when the pair is not a usable coordinate pair, the menu is shown at the element-relative default position, exactly as `$(el).contextMenu()` with no argument does. Reasoning:

* The reported trigger is a legitimate case, not a mistake. A keyboard activation genuinely has no pointer position, and the caller's intent is "show the menu on this element". Throwing turns a recoverable case into a crash on a code path the caller cannot always predict.
* It needs no new machinery. `position()` already routes to `determinePosition()` when it has no coordinates, and `$(el).contextMenu()` already means "position relative to the element". The fallback just leaves `pageX`/`pageY` unset on the triggered event and reuses that path, so there is one behaviour for "no coordinates" instead of two.
* The house style elsewhere in the plugin is to fall back for missing optional data (the `layerClick` `fakeClick` guard, the `e.data` guard in the contextmenu handler). It throws for structurally impossible input, such as a create without a selector, which this is not.

Details:

* A half-filled pair (`{x: 123, y: undefined}`) and a half-omitted one (`{x: 123}`) fall back too, rather than computing an offset of `NaN` or throwing.
* `0` is treated as a real coordinate everywhere. The default `position()` used to bail out to `determinePosition()` on `!x && !y`, which also caught the perfectly valid page origin. It now checks for actual numbers, with the `'maintain'` branch moved ahead of that check.
* Numeric strings (`{x: '123', y: '456'}`, e.g. read off a data attribute) still work, since the positioning arithmetic always accepted them. They are coerced before going on the event, so `e.pageX` is always a real number.

## Docs

`documentation/docs/plugin-commands.md` now states that `x` and `y` are **page** coordinates, the same space as `event.pageX` / `event.pageY` and scroll inclusive, not viewport coordinates and not relative to the trigger, with a short conversion snippet for viewport-based sources such as `getBoundingClientRect()`, a canvas or a map library. It also documents the fallback. See #726 for a worked example of why this matters.

## Backwards compatibility

No currently working call shape changes. `test/unit/issue-812-xy-overload.test.js` pins the ones that share this dispatch chain: no argument, `'hide'`, `'destroy'`, `false`/`true` (disable/enable), a plain definition object, a definition object that also carries `x`/`y`, and `{x: 123, y: 456}` reaching `position()` unchanged. Everything else this PR touches used to throw or to misposition, so the only behaviour changes are crash-to-working ones. Nothing new is thrown.

The one deliberate behaviour change is `{x: 0, y: 0}`: it used to be caught by the `!x && !y` bail-out and positioned relative to the element, and now lands at the page origin as documented. The existing `open contextMenu at 0,0` test still passes.

## Tests

New `test/unit/issue-812-xy-overload.test.js`: undefined coordinates no longer throw and fall back, an omitted `x` or `y` key falls back, finite and negative coordinates are forwarded unchanged, numeric strings arrive as numbers, `{x: 0, y: 0}` is honoured by both the routing and the default `position()`, a half-filled pair falls back without producing `NaN` css, non-numeric values fall back, a definition object carrying `x`/`y` still registers a menu, and the call shapes listed above are unchanged.

Full unit suite green (92 tests) and the 36 acceptance tests pass, since positioning is exercised there. `dist/` is intentionally untouched; `test/integration/html` is gitignored, so the docs edit brings no generated HTML with it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,12 +27,14 @@
 * A `select`'s phantom mousedown is no longer treated as an outside click (fixes #744)
 * A re-dispatched layer click now targets the element actually clicked (fixes #771)
 * Guard against undefined `e.data` in the contextmenu handler (fixes #777)
+* `$(...).contextMenu({x, y})` with missing or non-numeric coordinates now falls back to the element-relative position instead of throwing `No selector specified`, and an explicit `{x: 0, y: 0}` is honoured (fixes #812)
 
 #### Documentation
 
 * Documented using custom SVG icons without a gulp build step (fixes #762)
 * Added a dynamic per-row title example to the menu-title demo (fixes #769)
 * The asynchronous create demo now works on right click (fixes #735)
+* Documented that `$(...).contextMenu({x, y})` takes page coordinates (fixes #812)
 
 ### 2.10.2
 
--- a/documentation/docs/plugin-commands.md
+++ b/documentation/docs/plugin-commands.md
@@ -42,6 +42,18 @@ $(".some-selector").contextMenu();
 $(".some-selector").contextMenu({x: 123, y: 123});
 ```
 
+`x` and `y` are **page** coordinates, the same space as `event.pageX` / `event.pageY`, so they include the document scroll. They are not viewport coordinates and they are not relative to the trigger element. Coming from a viewport-based source (`event.clientX` / `event.clientY`, `getBoundingClientRect()`, a canvas or map library) add the current scroll offset:
+
+```
+var rect = element.getBoundingClientRect();
+$(".some-selector").contextMenu({
+  x: rect.left + window.scrollX,
+  y: rect.bottom + window.scrollY
+});
+```
+
+When either `x` or `y` is missing or is not a number, which happens when they are read off an event that carries no pointer position such as a keyboard or synthetic one, the menu falls back to `determinePosition` and is positioned relative to the trigger element, just like `$(".some-selector").contextMenu()`.
+
 ## Manually hide a contextMenu
 
 hide the contextMenu of the first element of the selector:
--- a/src/jquery.contextMenu.js
+++ b/src/jquery.contextMenu.js
@@ -2564,14 +2564,24 @@
     }
 
     // is the given `$.fn.contextMenu()` argument the {x, y} positioning
-    // overload rather than a menu definition? Decided on key *presence*: an
-    // event that carries no pointer position (a keyboard or synthetic one)
-    // yields {x: undefined, y: undefined}, which is still clearly meant as a
-    // position and must not be mistaken for a menu definition.
+    // overload rather than a menu definition? Decided on key *presence*, not on
+    // the values: an event that carries no pointer position (a keyboard or
+    // synthetic one) yields {x: undefined, y: undefined}, which is still
+    // clearly meant as a position and must not be mistaken for a menu
+    // definition. A menu definition wins whenever the object also carries a
+    // definition-only key, so an options object with an `x`/`y` of its own
+    // keeps being registered as a menu.
     // See https://github.com/swisnl/jQuery-contextMenu/issues/812
     function isCoordinateOperation(operation) {
-        return !!operation && typeof operation === 'object' &&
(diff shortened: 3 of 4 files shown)


bbrala:
Thanks, addressed in 1da51e9.

**Recognize single-key coordinate operations (P2) - agreed and fixed.** `{x: 123}` with the `y` key genuinely absent behaved differently from `{x: 123, y: undefined}`, which is exactly the inconsistency this PR set out to remove, and it contradicted what the docs and changelog now promise. The predicate now accepts *either* key.

Because that widens what counts as a positioning call, I paired it with a guard in the other direction: an object that also carries a definition-only key (`items`, `selector` or `build`) is still treated as a menu definition, so an options object that happens to have an `x`/`y` of its own cannot be hijacked into the positioning branch. There is no `x`/`y` *option* in the plugin, so this is belt and braces rather than a known case, but it keeps the widening strictly non-breaking.

New tests cover both: an omitted `x` and an omitted `y` key each fall back to `determinePosition`, and a definition object carrying `x`/`y` still registers a menu. I also pinned the neighbouring call shapes that share this dispatch chain - no argument, `'hide'`, `'destroy'`, `true`/`false`, and the plain definition object - so the widened detection is demonstrably non-breaking, and added negative coordinates alongside the existing `0` case.

92 unit tests and 36 acceptance tests green locally; CI was green on the first commit across jQuery 1.12.4 / 2.2.4 / 3.7.1 / 4.0.0.

